PR #4, opened 2018-12-20
fix: add types declaration to package.json

IlyaSemenov:
Currently types are not linked by `package.json` and will only be available if the environment recursively pulls all `*.d.ts` in `node_modules`, which is not always the case.

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,6 +8,7 @@
   ],
   "version": "1.0.1",
   "main": "dist/index.js",
+  "types": "dist/types.d.ts",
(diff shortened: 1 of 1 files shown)


IlyaSemenov:
Just in case, the current workaround is:

```ts
import { DeepPartial } from "ts-essentials/dist"
```

krzkaczor:
Thanks for your contribution 👍 That totally makes sense, I must have overlooked that.

krzkaczor:
Released as: `ts-essentials@1.0.2`

PR #6, opened 2019-01-26
Add Merge type

Andarist:


krzkaczor:
That looks cool, thanks! Could you add readme note as well (its manual work 😐)

PR #7, opened 2019-01-29
Add ValueOf type

macbem:
This PR adds a type which extracts types from object values - nothing ground breaking, it just makes commonly written typespecs a bit more concise.


krzkaczor:
Wow, Maciek! Great Moves, Keep It Up, Proud of You! 😆Thanks!

Released as `ts-essentials@1.0.4`.

PR #10, opened 2019-02-05
Add Writable type

Andarist:


krzkaczor:
Thanks! Could you add it to README as well? 🙏 

Btw. i fixed CI setttings so it should run on forked repos as well.

Andarist:
> Thanks! Could you add it to README as well? 🙏

Done.

krzkaczor:
Released as `ts-essentials@2.0.0`

PR #11, opened 2019-02-05
Remove custom NonNullable

Andarist:
⚠️ breaking change

I might miss something, but can't find a reason behind this custom implementation. If there is any I would appreciate an explanation why it's needed.


krzkaczor:
You are totally right. I wasn't aware that it's part of standard lib.

krzkaczor:
Released as `ts-essentials@2.0.0`

PR #18, opened 2019-02-15
Add ReadonlyArray<T> check in DeepReadonly

MichaelTontchev:
As per https://github.com/krzkaczor/ts-essentials/issues/17, DeepReadonly wasn't properly recognizing a ReadonlyArray\<T> as matching any[], with the result being that such fields were wrapped in an extra DeepReadonlyObject type, which caused type clashing elsewhere. See the issue for a repro.

### DIFF
--- a/lib/types.ts
+++ b/lib/types.ts
@@ -17,7 +17,7 @@ export type DeepPartial<T> = {
 /** Like Required but recursive */
 export type DeepRequired<T> = T extends Primitive
   ? NonNullable<T>
-  : T extends any[]
+  : T extends (any[] | ReadonlyArray<any>)
   ? DeepRequiredArray<NonNullable<T[number]>>
   : T extends {}
   ? { [K in keyof T]-?: DeepRequired<NonNullable<T[K]>> }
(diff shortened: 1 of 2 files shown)


lib/types.ts
```diff
@@ -17,7 +17,7 @@ export type DeepPartial<T> = {
 /** Like Required but recursive */
 export type DeepRequired<T> = T extends Primitive
   ? NonNullable<T>
-  : T extends any[]
+  : T extends (any[] | ReadonlyArray<any>)
```
krzkaczor: isn't ReadonlyArray<any> good enough?

MichaelTontchev: Technically it is, but I think this way is better for two reasons:
1) It's more explicit code: it's unclear until you think for some time why a normal array can be matched by ReadonlyArray\<any>. Most people reading this code will wonder "where are they matching normal arrays?"
2) It's more future-proof. If in the future any[] and ReadonlyArray\<any> don't match the same set of types, this construction will ensure both are caught.

MichaelTontchev: Aaaand just realized I had made the change in the wrong place :| Fixed it.

MichaelTontchev:
@krzkaczor updated the PR with a test case

krzkaczor:
@MichaelTontchev thanks for your contribution! 🎉

krzkaczor:
Released as `2.0.1`.

PR #21, opened 2019-02-25
Add ClearNever

Andarist:


krzkaczor:
What do you think about naming this similarly to `NonNullable` built in? Maybe `NonNever` :D I think it could make it simpler to understand the purpose of this type. Could you leave one line explanation in the readme as well? 

Also, FYI: there's a possibility to write some [tests](https://github.com/krzkaczor/ts-essentials/blob/master/test/index.ts) using type assertion library. This should be fairly easy to test.

Andarist:
Personally I like `ClearNever` better, but I understand that `NonNever` would relate to `NonNullable`. Gonna change in a moment.

> Also, FYI: there's a possibility to write some [tests](https://github.com/krzkaczor/ts-essentials/blob/master/test/index.ts) using type assertion library. This should be fairly easy to test.

I know, I was too lazy to write tests though :trollface: Gonna add some in a moment.

Andarist:
Just to reiterate on the naming - `NonNullable` operates on basic types, `NonNever` operates on object types and clears properties which have never value.

krzkaczor:
@Andarist yeah, I have also realized this difference. I am not pushing renaming it really hard, I just think it's slightly easier to understand now 😆 

Thanks for writing the test 👍

PR #25, opened 2019-03-08
Make Tuple parametrized with generic so it can be used for granular c…

Andarist:
…onditional checks

### DIFF
--- a/README.md
+++ b/README.md
@@ -205,6 +205,8 @@ const ret = foo(["s", 1]);
 
 ```
 
+You can also parametrize `Tuple` type with a type argument to constraint it to certain types, i.e. `Tuple<string | number>`.
+
 ### Literal types
 
 ```typescript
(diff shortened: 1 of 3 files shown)


Andarist:
After a though - I'm not sure if this works correctly. It has fixed one of my issues but not sure if for good reasons. Generics are hard 🙈

krzkaczor:
@Andarist Did you confirm that it works correctly? Tmrw I will conduct some tests on my own.

Andarist:
I think it does, gonna try to write some test for this in a moment.

Andarist:
Added test - it correctly restricts input tuples. It also accepts empty array though which is kinda a bummer.

krzkaczor:
@Andarist ok cool, thanks! One last thing: could leave a note about parameters in readme?

Andarist:
Done

krzkaczor:
@Andarist thanks! 👍

PR #31, opened 2019-04-18
fix(package.json): function definition not exported

yxliang01:
Before this fix, only type definitions are exported. This fixes https://github.com/krzkaczor/ts-essentials/issues/26 .

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
   ],
   "version": "2.0.4",
   "main": "dist/index.js",
-  "types": "dist/types.d.ts",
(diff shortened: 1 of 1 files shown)


krzkaczor:
OMG, 🤦‍♂️ great finding! I was also wondering what's really causing these issues. Thanks!

krzkaczor:
Btw. don't worry about failing CI. I am playing with this new project — codechecks, it provides additional information during code reviews but it doesn't work with forks *yet*.

PR #33, opened 2019-04-25
test: add test testDeepWritableReverseIsDeepReadonlyForTotallyWritableType

yxliang01:
Test whether for totally writable types, apply `DeepReadonly` then `DeepWritable` will yield the original type. This is expected to be true for all totally writable types.

### DIFF
--- a/test/index.ts
+++ b/test/index.ts
@@ -137,3 +137,16 @@ function testDeepWritable2() {
   test[0].foo = "b";
   test[0].bar.x = 2;
 }
+
+// Test whether for totally writable types, apply DeepReadonly then DeepWritable will yield the original type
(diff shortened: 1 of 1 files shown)


yxliang01:
@krzkaczor Hmm the CI failure doesn't look related to the new test. Same as the problem described in https://github.com/krzkaczor/ts-essentials/pull/31#issuecomment-484674688 ?

krzkaczor:
I love the name of this test 😆 

Yes, it's the same problem. Don't worry about it, it will be fixed soon.

krzkaczor:
Thanks for your contribution!

yxliang01:
Very long name though. :P

PR #36, opened 2019-04-25
docs(README.md): add backport words for literal

yxliang01:


krzkaczor:
Yeah, I think it makes sense 👍 Could you rebase with master? I updated codechecks so now it shouldn't fail.

yxliang01:
@krzkaczor Rebased.

krzkaczor:
Okay I see that it still has some problems 😆 Anyway, codechecks provides additional checks in github interface. In this project we track build size (both for d.ts and js files) This is pretty useful to see what impact changes have on existing code.

krzkaczor:
Thanks! Merged

yxliang01:
> Okay I see that it still has some problems  Anyway, codechecks provides additional checks in github interface. In this project we track build size (both for d.ts and js files) This is pretty useful to see what impact changes have on existing code.

Maybe it's a bit off. But, I am curious why the build size matters much? I can understand if this is to be shipped to browsers. But this is not as this is a build-time use library.

krzkaczor:
The dirty secret of this lib is that actually **some** types are part of the bundle ([functions](https://github.com/krzkaczor/ts-essentials/blob/master/lib/functions.ts) file). I need to note this in readme.

So, this breakdown makes sense: https://github.com/krzkaczor/ts-essentials/pull/36/checks since you can track exact code that would become part of the bundle. 

Codechecks.io generally allow you to write any logic that is able to track changes across PRs, if you're interested you can take a look at docs: https://github.com/codechecks/docs (it's still in private beta, let me know if you want to get in ;) ).

PR #41, opened 2019-05-01
Fix DeepReadonly<unknown>

ThomasdenH:
This commit makes `unknown` assignable to `DeepReadonly<unknown>`.

### DIFF
--- a/lib/types.ts
+++ b/lib/types.ts
@@ -31,7 +31,9 @@ export type DeepReadonly<T> = T extends Primitive
   ? DeepReadonlyArray<T[number]>
   : T extends Function
   ? T
-  : DeepReadonlyObject<T>;
+  : T extends {}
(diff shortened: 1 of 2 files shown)


krzkaczor:
Nice PR! I wonder if changing the order of conditions make sense but I will test it on my own. Thanks!

krzkaczor:
Released as `2.0.7`

ThomasdenH:
Great! The reason for the order is that anything `extends unknown`.

krzkaczor:
Yeah, I realized that as well after I played with it for a little bit (I also saw your previous attempts for extending `Primitive` type). 

I think your solution fine 👍 

Btw. other `deep` types probably need a similiar fix. @ThomasdenH if you want to fix it just go ahead ;)

PR #56, opened 2019-06-10
Add MarkRequired type

### LINKED ISSUE #45: Proposal: PickRequired / PartialRequired
I'm using this generic type which I believe is a good fit for ts-essentials:

```ts
export type PickRequired<T, K extends keyof T> = { [P in K]-?: T[P] }
```

**UPDATE 10.06.2019: The suggested implementation is not correct, as it does not say anything about keys *not* in `T`, and essentially repeats `Pick<Required<..>>`. See the discussion below.**

Example usage:

```ts
const user = (await User.query()
	.select("name", "is_admin")
	.where("id", user_id)
	.first()) as PickRequired<User, "name" | "is_admin">
```

Normally, ORM query would return a `Partial<...>` type, because it's not clear which fields will be present in the result set. With `PickRequired`, I can statically attribute that certain fields **will** be present in the result.

I'm not exactly sure about the naming. I considered `PartialRequired` and `PickRequired`, but both bear the unwanted connotations to existing Typescript generics (`Partial` and `Pick`). Pick seemed to be better as it similarly accepts the list of keys.

**UPDATE 10.06.2019: Two more naming variants are `MarkRequired` and `SelectiveRequired`.**

quezak:
Closes #45 

Named the type `MarkRequired` after discussion.

### DIFF
--- a/lib/types.ts
+++ b/lib/types.ts
@@ -63,6 +63,9 @@ export type NonNever<T extends {}> = Pick<T, { [K in keyof T]: T[K] extends neve
 /** Merge 2 types, properties types from the latter override the ones defined on the former type */
 export type Merge<M, N> = Omit<M, Extract<keyof M, keyof N>> & N;
 
+/** Mark some properties as required, leaving others unchanged */
+export type MarkRequired<T, RK extends keyof T> = Exclude<T, RK> & Required<Pick<T, RK>>;
+
 /** Convert union type to intersection #darkmagic */
 export type UnionToIntersection<U> = (U extends any ? (k: U) => void : never) extends ((k: infer I) => void)
   ? I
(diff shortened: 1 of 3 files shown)


lib/types.ts
```diff
@@ -63,6 +63,9 @@ export type NonNever<T extends {}> = Pick<T, { [K in keyof T]: T[K] extends neve
 /** Merge 2 types, properties types from the latter override the ones defined on the former type */
 export type Merge<M, N> = Omit<M, Extract<keyof M, keyof N>> & N;
 
+/** Mark some properties as required, leaving others unchanged */
+export type MarkRequired<T, RK extends keyof T> = Exclude<T, RK> & Required<Pick<T, RK>>;
```
krzkaczor: NOTE: i think the other way to implement is to use `Merge` type defined above. `Merge<T, Required<Pick<T, RK>>>` just as a side note

quezak: I've actually used `Merge` here in the beginning, but then the created types grow overcomplicated (not really readable when you hover over them in vscode). Because this merges disjoint parts of the same type, we don't have to explicitly setup overrides, and the output is a bit cleaner.

PR #59, opened 2019-06-12
Add WritableKeys & ReadonlyKeys types

Andarist:


lib/types.ts
```diff
@@ -82,3 +82,16 @@ export type Tuple<T = any> = [T] | T[];
 
 /** Useful as a return type in interfaces or abstract classes with missing implementation */
 export type AsyncOrSync<T> = PromiseLike<T> | T;
+
+/** Dark magic helper, needed for WritableKeys & ReadonlyKeys */
+type IfEquals<X, Y, A = X, B = never> = (<T>() => T extends X ? 1 : 2) extends (<T>() => T extends Y ? 1 : 2) ? A : B;
+
+/** Gets keys of an object which are readonly */
+export type ReadonlyKeys<T extends object> = {
```
krzkaczor: woah these look complicated. Could you write some tests? I will need a while to understand them

krzkaczor: any tips helping us to undestand this? code comments would be appreciated

Andarist: Truth to be told I don't fully understand this either, no way I could have written on my own. Found these on SO while searching how I could achieve those.

The one thing that we may notice that implementations of WritableKeys and ReadonlyKeys are very similar, the only difference is 3rd and 4th type argument given to #darkmagic IfEquals. This makes sense as writable and readonly keys are indeed complementary sets.

The weirdest thing happens in the implementation of `IfEquals` - we check if a subtype of a function returning "left" type is a subtype of a subtype of a function returning "right" type and based on that left or bad is chosen.

I don't even know why I've tried to explain this here, I don't understand, this shit is crazy, I feel dumb, my head hurts, I need a drink...

krzkaczor:
Thanks! I am gonna write some additional docs and release it.

krzkaczor:
Released as `2.0.10`

PR #62, opened 2019-06-24
fix: check Primitive in DeepPartial (#19)

### LINKED ISSUE #19: DeepPartial does not work with index signatures
The following errors:

```ts
import { DeepPartial } from 'ts-essentials';

interface Params {
  [key: string]: any
}

interface Route {
  params: Params
}

const route: DeepPartial<Route> = {
  params: {
    test: '123'
  }
}
```

```
Type 'string' is not assignable to type 'DeepPartial<{}>[] | ReadonlyArray<DeepPartial<{}>> | DeepPartial<any>'.
```

https://stackblitz.com/edit/typescript-eu3ajv

lucifer1004:
This PR should fix #19

### DIFF
--- a/test/index.ts
+++ b/test/index.ts
@@ -25,6 +25,7 @@ function testDeepPartial() {
     simple: number;
     nested: {
       date: Date;
+      func: () => string;
     };
     params: Params;
   };
@@ -32,9 +33,12 @@ function testDeepPartial() {
   type Expected = {
     simple?: number;
     nested?: {
-      date?: DeepPartial<Date>;
+      date?: Date;
+      func?: () => string;
     };
-    params?: DeepPartial<Params>;
+    params?: {
+      [key: string]: any;
+    };
   };
 
   type Test = Assert<IsExact<DeepPartial<Input>, Expected>>;
--- a/test/index.ts
+++ b/test/index.ts
@@ -18,24 +18,24 @@ import {
 
 function testDeepPartial() {
   type Params = {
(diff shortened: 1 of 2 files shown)


krzkaczor:
Could you write some tests? I think there are already some cases for DeepPartial. Thanks!

test/index.ts
```diff
@@ -15,6 +16,30 @@ import {
...
+    nested: {
+      date: Date;
+    };
+    params: Params;
+  };
+
+  type Expected = {
+    simple?: number;
+    nested?: {
+      date?: DeepPartial<Date>;
```
krzkaczor: Do not use `DeepPartial` in expected type since its still not clear what should be the result.
Probably the type should look like this:

```typescript
type Expected = {
    simple?: number;
    nested?: {
      date?: Date;
    };
    params?: {
      [key: string]?: any;
    };
  };
```

lucifer1004: `[key: string]?: any` is not valid, and I have not come up with a solution.

lucifer1004: Since `[key: string]` is already a partial property, I think we can directly use `[key: string]: any`

krzkaczor:
Solves https://github.com/krzkaczor/ts-essentials/issues/37 as well.

krzkaczor:
@lucifer1004 Great PR, thanks! Merging it now.

krzkaczor:
Published as `ts-essentials@2.0.11`

Issue #63, opened 2019-06-24
Extending definition of Primitive?


lucifer1004:
Referring to [type-fest](https://github.com/sindresorhus/type-fest/blob/89f09b3133beb637f3869ba827e1c909acc26c6b/source/basic.d.ts#L7), `bigint`and `symbol` can be added.

Also, it should be noted that

> This can just be `export type Primitive = not object` when the `not` keyword is out.

krzkaczor:
Adding `bigint` and `symbol` definately make sense. 

Negated types (`not`) are gonna help solving https://github.com/krzkaczor/ts-essentials/issues/35 as well

krzkaczor:
@all-contributors please add @lucifer1004 for code, idea

krzkaczor:
Fixed by https://github.com/krzkaczor/ts-essentials/pull/65

PR #65, opened 2019-06-24
feat: add bigint and symbol to Primitive (#63)

lucifer1004:
This PR is as the title describes.


krzkaczor:
Thanks! Published as `ts-essentials@2.0.11`

Issue #40, opened 2019-04-29
DeepReadonly on Tuple type causes type widening to Array


yxliang01:
```typescript
import { DeepReadonly } from "ts-essentials";

const a: [number, string] = [1,'s'];
const b: DeepReadonly<typeof a> = a;
const c = b[1];
```

The variable `c` above will have type `number | string` while the expected type would be `string`. This is caused by the fact that the Tuple type is widened to Array type.

krzkaczor:
Part of `next` branch. Released as `ts-essentials@3.0.0-beta.0`

Issue #14, opened 2019-02-07
Add DeepNonNullable type


krzkaczor:

krzkaczor:
Part of `next` branch. Released as `ts-essentials@3.0.0-beta.0`

Issue #12, opened 2019-02-07
Add generic Deep type


krzkaczor:
Is it possible to create generic Deep type that would minimize repetition between `DeepPartial`, `DeepReadonly` etc

Andarist:
Was thinking about it a little bit and I think this aint possible - although Im quite new to TS so I might miss something.

The challenge here is that each of those types does something different to property modifiers, and also we don't know how to map a property modifier to array modifier. So at the very least we would require 3 type parameters `<RecursiveMapper, ObjectHandler, ArrayHandler>` which seems more than convoluted and I still don't know how to apply `ObjectHandler` to the transformation.

Beraliv:
You want to have high order function in TypeScript (which is discussed here – https://github.com/microsoft/TypeScript/issues/1213)

You still can define `Replace` and change one type with another:

```ts
declare const __placeholder__: unique symbol;

type Placeholder = { readonly [__placeholder__]: unknown };

type Replace<T, X, Y> = {
  [K in keyof T]: T[K] extends X ? Y : T[K];
};
```

But let me check that it can be implemented somehow

Beraliv:
`Replace` has a lot of limitations, e.g.

1. It isn't working if you don't know element types in advance – https://tsplay.dev/W4pbXW
2. Can use `Replace` for one type of parameter, but if you have multiple placeholders, you need to have more and follow it really carefully
3. You can `Replace` only on first level, or implement a deep replacement (might be difficult)

Also, you might need several implementation of `Replace`: for objects and primitives

Beraliv:
Several days ago I've seen work on hot script - https://github.com/gvergnaud/hotscript

Maybe it will be possible to organise it so it would be possible to reuse the same types, although I didn't have a chance to play with it

Beraliv:
I've closed the PR because it is not a priority right now

PR #66, opened 2019-06-24
A complete refactor to Deep* types

lucifer1004:
- Refactor DeepReadonly, DeepWritable, DeepPartial
- Implement DeepNonNullably
- Refactor and add related tests

This refactor should solve #14 #40 
And since all Deep types are very similar now, it provides a foundation for #12

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@ yarn add --dev ts-essentials
 - [Dictionaries](#dictionaries)
   - Dictionary
   - DictionaryValues
-- [Deep Partial & DeepRequired & Deep Readonly](#deep-partial--deep-required--deep-readonly)
+- [Deep Partial & DeepRequired & Deep Readonly & Deep NonNullablee](#deep-partial--deep-required--deep-readonly--deep-nonnullable)
 - [Writable & DeepWritable](#writable)
 - [Omit](#omit)
 - [OmitProperties](#omitproperties)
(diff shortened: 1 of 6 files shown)


lucifer1004:
VS Code yields no errors, however, tsc fails.

lucifer1004:
It is related to prettier... If I do not format the code, tsc can pass.

lucifer1004:
TypeScript > 3.5 is OK, but 3.2 fails.

lucifer1004:
I suggest we follow TypeScript > 3.5, in which arrays do not need specific judgment, and can be treated just like {}

lucifer1004:
By treating arrays just like {}, issues with tuples (#40) are also addressed like charm.

lucifer1004:
I realized that the easiest way is to upgrade TypeScript version in devDeps. So it's up to @krzkaczor to decide whether it is time to perform the upgrade.

krzkaczor:
This is a really enormous help for the project! Thanks for this PR!

Bumping TS version is for sure not perfect but on the other hand, we can release this as version `ts-essentials@3.0.0` and let folks know that they can still use 2.x.x if they use older TS version.

I will need a while to review this PR. Hopefully today/tmrw I will get right back to you.

krzkaczor:
Thanks! I am gonna merge this 🎉 

I created a new branch `next` and that is the branch for explicit support for TS 3.5. I will also publish it as `ts-essentials@3.0.0-beta.1` so it's already usable for early adopters. My plan would be to wait a little bit before releasing it to master to give the community time to migrate to TS 3.5. On `next` branch we can also remove deprecated types that became part of TS standard lib (`omit`)

krzkaczor:
Part of `next` branch. Released as `ts-essentials@3.0.0-beta.0`

PR #67, opened 2019-06-25
Implement DeepGeneric

lucifer1004:
This PR implements DeepGeneric on basis of #66. 
The implementation is far from elegant, so maybe it should not be merged.
@krzkaczor Waiting for your reply.

### DIFF
--- a/lib/types.ts
+++ b/lib/types.ts
@@ -5,93 +5,63 @@ export type Primitive = string | number | boolean | bigint | symbol | undefined 
 export type Dictionary<T, K extends string | number = string> = { [key in K]: T };
 export type DictionaryValues<T> = T extends Dictionary<infer U> ? U : never;
 
-/** Like Partial but recursive */
-export type DeepPartial<T> = T extends Primitive
-  ? T
-  : T extends Function
-  ? T
-  : T extends Date
-  ? T
-  : T extends Map<infer K, infer V>
-  ? DeepPartialMap<K, V>
-  : T extends Set<infer U>
-  ? DeepPartialSet<U>
-  : T extends {}
-  ? { [K in keyof T]?: DeepPartial<T[K]> }
-  : Partial<T>;
-interface DeepPartialSet<ItemType> extends Set<DeepPartial<ItemType>> {}
-interface DeepPartialMap<KeyType, ValueType> extends Map<DeepPartial<KeyType>, DeepPartial<ValueType>> {}
-
-/** Like NonNullable but recursive */
-export type DeepNonNullable<T> = T extends Primitive
-  ? NonNullable<T>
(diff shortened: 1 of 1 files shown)


lib/types.ts
```diff
@@ -5,93 +5,63 @@ export type Primitive = string | number | boolean | bigint | symbol | undefined
...
 export type DictionaryValues<T> = T extends Dictionary<infer U> ? U : never;
 
-/** Like Partial but recursive */
-export type DeepPartial<T> = T extends Primitive
-  ? T
-  : T extends Function
-  ? T
-  : T extends Date
-  ? T
+type NonNullableSymbol = "NonNullable";
```
krzkaczor: smart, I think there is no better way to implement this.

krzkaczor: @quezak @Andarist @macbem any ideas?

krzkaczor: Just to sum up, @lucifer1004 used symbols to manually invoke correct generic type later. I think it's not possible to pass "not-applied generic type as a type parameter" to avoid that...

Andarist: Not sure how I feel about it - seems like the implementation got rather complex for arguable gains.

lucifer1004: @krzkaczor Once we could use generic type as a type parameter, things would be a lot easier.

lucifer1004: > Not sure how I feel about it - seems like the implementation got rather complex for arguable gains.

@Andarist You are right. That's why I don't think this PR is ready to be merged. Anyway, the implementation is ugly and not easily extensible. If we want to add a new Deep type later, the definition of DeepGeneric has to be modified, and that is apparently not what we want.

krzkaczor: > @krzkaczor Once we could use generic type as a type parameter, things would be a lot easier.

Did you find any issue in TS repository about this? I am not sure how it should be phrased correctly. 

> @Andarist You are right. That's why I don't think this PR is ready to be merged. Anyway, the implementation is ugly and not easily extensible. If we want to add a new Deep type later, the definition of DeepGeneric has to be modified, and that is apparently not what we want.

Yeah, definitely this is not perfect :/ I think for now we shouldn't merge this.

quezak: > Not sure how I feel about it - seems like the implementation got rather complex for arguable gains

I agree -- while it would be nice to add a "Deep Anything" abstraction, the current implementation is indeed kinda _ugly and not easily extensible_, as the author said himself. We can keep this PR open for future reference, in case someone comes up with a cleaner idea -- but until then, IMO the `Deep*` types should remain in the current form, that is a bit easier to read.

krzkaczor:
Closing it for now as we need a nicer way to implement this.

PR #68, opened 2019-06-28
Implement DeepOmit

lucifer1004:
Below is a simple example (which has been added to README)

Here is the `Teacher` interface.

```typescript
interface Teacher {
  name: string,
  gender: string,
  students: {name: string, score: number}[]
}
```

Now suppose you want to omit `gender` property of `Teacher`, and `score` property of `students`. You can simple define a filter type and then use `DeepOmit`.

In the filter type, the properties to be omitted completely should be defined as `true`. For the properties you want to partially omit, you should recursively define the sub-properties to be omitted.

```typescript
interface Filter {
  gender: true,
  students: {
    score: true,
  }
}

type TeacherSimple = DeepOmit<Teacher, Filter>

// The result will be:
// {
//  name: string,
//  students: {name: string}[]
// }
```

NOTE: `DeepOmit` works fine with `Array`s and `Set`s. When applied to a  `Map`, if its key and value shares common properties, both will be omitted.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,32 +25,32 @@ yarn add --dev ts-essentials
 
 ## What's inside?
 
-- [Install](#install)
-- [What's inside?](#whats-inside)
-  - [Basic:](#basic)
-  - [Dictionaries](#dictionaries)
-  - [Deep Partial & Deep Required & Deep Readonly & Deep NonNullable](#deep-partial--deep-required--deep-readonly--deep-nonnullable)
-  - [Writable](#writable)
-  - [Omit](#omit)
-  - [StrictOmit](#strictomit)
-    - [Comparison between `Omit` and `StrictOmit`](#comparison-between-omit-and-strictomit)
-  - [DeepOmit](#deepomit)
-  - [OmitProperties](#omitproperties)
-  - [NonNever](#nonnever)
-  - [Merge](#merge)
-  - [MarkRequired](#markrequired)
-  - [ReadonlyKeys](#readonlykeys)
-  - [WritableKeys](#writablekeys)
-  - [UnionToIntersection](#uniontointersection)
-  - [Opaque types](#opaque-types)
-  - [Tuple constraint](#tuple-constraint)
-  - [Literal types](#literal-types)
-  - [Exhaustive switch cases](#exhaustive-switch-cases)
-  - [ValueOf type](#valueof-type)
-  - [AsyncOrSync type](#asyncorsync-type)
-- [Contributors](#contributors)
+- [Install](#Install)
+- [What's inside?](#Whats-inside)
+  - [Basic](#Basic)
+  - [Dictionaries](#Dictionaries)
+  - [Deep Partial & Deep Required & Deep Readonly & Deep NonNullable](#Deep-Partial--Deep-Required--Deep-Readonly--Deep-NonNullable)
+  - [Writable](#Writable)
+  - [Omit](#Omit)
+  - [StrictOmit](#StrictOmit)
+    - [Comparison between `Omit` and `StrictOmit`](#Comparison-between-Omit-and-StrictOmit)
+  - [DeepOmit](#DeepOmit)
+  - [OmitProperties](#OmitProperties)
+  - [NonNever](#NonNever)
+  - [Merge](#Merge)
+  - [MarkRequired](#MarkRequired)
+  - [ReadonlyKeys](#ReadonlyKeys)
+  - [WritableKeys](#WritableKeys)
+  - [UnionToIntersection](#UnionToIntersection)
+  - [Opaque types](#Opaque-types)
+  - [Tuple constraint](#Tuple-constraint)
+  - [Literal types](#Literal-types)
+  - [Exhaustive switch cases](#Exhaustive-switch-cases)
+  - [ValueOf type](#ValueOf-type)
+  - [AsyncOrSync type](#AsyncOrSync-type)
+- [Contributors](#Contributors)
 
-### Basic:
+### Basic
 
 - `Primitive` type matching all primitive values.
 
@@ -228,13 +228,13 @@ interface Teacher {
 
 Now suppose you want to omit `gender` property of `Teacher`, and `score` property of `students`. You can achieve this with a simple type filter.
 
-In the filter, the properties to be omitted completely should be defined as `true`. For the properties you want to partially omit, you should recursively define the sub-properties to be omitted.
+In the filter, the properties to be omitted completely should be defined as `never`. For the properties you want to partially omit, you should recursively define the sub-properties to be omitted.
 
 ```typescript
 type TeacherSimple = DeepOmit<Teacher, {
-  gender: true,
+  gender: never,
   students: {
-    score: true,
+    score: never,
   }
 }>
 
@@ -245,7 +245,8 @@ type TeacherSimple = DeepOmit<Teacher, {
 // }
 ```
 
-NOTE: 
+NOTE
+
 - `DeepOmit` works fine with `Array`s and `Set`s. When applied to a `Map`, the filter is only applied to its value.
 - If there exists any property in the filter which is not in the original type, an error will occur.
 
--- a/README.md
+++ b/README.md
@@ -32,6 +32,7 @@ yarn add --dev ts-essentials
   - [Deep Partial & Deep Required & Deep Readonly & Deep NonNullable](#Deep-Partial--Deep-Required--Deep-Readonly--Deep-NonNullable)
   - [Writable](#Writable)
   - [Omit](#Omit)
+  - [DeepOmit](#DeepOmit)
   - [OmitProperties](#OmitProperties)
   - [NonNever](#NonNever)
   - [Merge](#Merge)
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -162,31 +163,47 @@ test[0].bar.x = 2;
...
+Now suppose you want to omit `gender` property of `Teacher`, and `score` property of `students`. You can simple define a filter type and then use `DeepOmit`.
 
-// Result:
-// {
-//  simple: number
-// }
+In the filter type, the properties to be omitted completely should be defined as `true`. For the properties you want to partially omit, you should recursively define the sub-properties to be omitted.
+
+```typescript
+interface Filter {
```
krzkaczor: what do you think about inlining filter? I guess this is what users would rather do because Filter type is not useful anywhere else.

quezak: +1, in the example usage it's enough to inline it

lib/types.ts
```diff
@@ -93,15 +93,26 @@ export type DeepWritable<T> = T extends Primitive
...
+  ? T
+  : T extends Map<infer K, infer V>
+  ? DeepOmitMap<K, V, P>
+  : T extends Set<infer U>
+  ? DeepOmitSet<U, P>
+  : T extends Array<infer U>
+  ? DeepOmitArray<U, P>
+  : T extends {}
+  ? { [K in Exclude<keyof T, keyof P>]: T[K] } &
+      OmitProperties<{ [K in Extract<keyof T, keyof P>]: P[K] extends true ? never : DeepOmit<T[K], P[K]> }, never>
```
quezak: Nice trick with the `OmitProperties<..., never>`, I was first extracting the key set and then mapping only the required props

README.md
lucifer1004: I explicitly defined the Filter type here to make it clearer. However, if you think inlining is better, I can use inlining filter in the README instead.

lib/types.ts
```diff
@@ -93,15 +93,26 @@ export type DeepWritable<T> = T extends Primitive
...
+  : T;
+interface DeepOmitArray<ItemType, P> extends Array<DeepOmit<ItemType, P>> {}
+interface DeepOmitSet<ItemType, P> extends Set<DeepOmit<ItemType, P>> {}
+interface DeepOmitMap<KeyType, ValueType, P> extends Map<DeepOmit<KeyType, P>, DeepOmit<ValueType, P>> {}
```
quezak: I don't like the fact that the same type filter is applied to both keys and values.
Maybe filter only values? Indexing maps with objects is probably a rare case.
Maybe write the type filter for `Map` separately for keys and values?

Or a combination of the above: using just one type in the filter filters values, if you want to filter keys too, use `Map<KeyFilter, ValueFilter>`:

```typescript
type DeepOmitMap<KeyType, ValType, Filter> =
  Filter extends Map<infer KeyFilter, infer ValFilter>
    ? Map<DeepOmit<KeyType, KeyFilter>, DeepOmit<ValType, ValFilter>>
    : Map<KeyType, DeepOmit<ValType, Filter>>
```

This looks nice, but maybe I'm going too far for the sake of type purity? :joy_cat: Do you have any ideas to avoid using the same type filter for keys & values?

quezak: CC @macbem

krzkaczor:
> The major difference from my proposal is that your filter type can have arbitrary keys, while mine was restricted to use only the actual keys of the filtered type. This can be helpful to avoid type typos (hehe), but maybe it's overkill?

I think we should be strict and allow only types that have common keys with a type being filtered — I really liked this in your original solution @quezak I can't imagine a situation when not finding such mistake makes sense.

quezak:
> I think we should be strict and allow only types that have common keys with a type being filtered 

Ok :) I think my type can be incorporated into this PR and extended for work with Arrays, Sets and Maps. Let's wait for @lucifer1004, or I could look into it later.

Maybe we should even restore "our better Omit" as `StrictOmit` or something :joy:

krzkaczor:
> Maybe we should even restore "our better Omit" as StrictOmit or something 😂

Yeah I didnt realize that there is a difference in their implementation. Our type is so much better 😆

lucifer1004:
@quezak I have tried your solution but found a problem.

`Spec extends DeepOmitSpec<T>` cannot force common types, because

`{a: string, b: number} extends {a: string}` is `true`.

So even with the spec, filters with nonexistent properties still pass.

lib/types.ts
lucifer1004: For now, I prefer to only apply the filter to the value of `Map`.

lucifer1004:
@krzkaczor I have added the filter spec suggested by @quezak 
It took quite a while to handle the Array/Set/Map case.

For Map, I now only filter V instead of K+V.

The problem is described in the comment above.

lucifer1004:
I come up with a new idea, wait for my update.

lucifer1004:
@krzkaczor @quezak 
Please check my latest implementation.

My idea is that T should be the **subset** of a **superset** of P.

This implementation ensures that neither wrong property type nor extra properties are allowed in the filter.

quezak:
Looks promising, I'll take a detailed look later to compare the two approaches.

lucifer1004:
Tried my best but failed to figure out a way to support immediately nested Set/Map/Array.

Works fine for things like:

```ts
type T = Set<{
    a: {
      b: string;
      c: {
        d: string;
        e: boolean;
      };
      f: number;
    };
    array: { a: string; b: boolean }[];
    set: Set<{ a: string; b: boolean }>;
    map: Map<number, { a: string; b: boolean }>;
  }>
```

But not this

```ts
type T = Set<{
    a: {
      b: string;
      c: {
        d: string;
        e: boolean;
      };
      f: number;
    };
    array: { a: string; b: boolean }[];
    set: Set<{ a: string; b: boolean }>;
    map: Map<number, { a: string; b: boolean }>;
  }>[]
```

nor this:

```ts
type T = {
    a: {
      b: string;
      c: {
        d: string;
        e: boolean;
      };
      f: number;
    };
    array: { a: string; b: boolean }[];
    set: Set<{ a: string; b: boolean }>;
    map: Map<number, Set<{ a: string; b: boolean }>>;
  }
```

lucifer1004:
I just checked and confirmed that my original version could support such nested omit.

A practical use case would be deep omitting in a high-dimensional array.

So there is a tradeoff.

lucifer1004:
To sum up:

|                          | Commit  | Cascaded Array/Set/Map | Security                              |
| ------------------------ | ------- | ---------------------- | ------------------------------------- |
| Impl 1 Plain             | 30da8b9 | ✔️                      | No type check                         |
| Impl 2 Filter Spec       | d486615 | ❎                      | Check wrong types                     |
| Impl 3 Superset + Subset | 813dbf1 | ❎                      | Check wrong types & nonexistent types |

lucifer1004:
Cascaded Array problem solved via the latest commit.

@krzkaczor @quezak Looking forward to your comments.

lib/types.ts
```diff
@@ -93,15 +93,49 @@ export type DeepWritable<T> = T extends Primitive
...
- */
-export type Omit<T, K extends keyof T> = Pick<T, Exclude<keyof T, K>>;
+/** Similar to the builtin Omit, but checks the filter strictly. */
+export type StrictOmit<T, K extends keyof T> = Pick<T, Exclude<keyof T, K>>;
 
 /** Omit all properties of given type in object type */
 export type OmitProperties<T, P> = Pick<T, { [K in keyof T]: T[K] extends P ? never : K }[keyof T]>;
 
+/** Recursively omit deep properties */
+export type DeepOmit<T extends DeepOmitModify<P>, P> = T extends Primitive | Function | Date
```
quezak: `T` and `K` are ok, but could you use names for the other generic params? Like `Filter` instead of `P`, `Map<infer KeyT, infer ValT>`, `Array<infer ElemT>`, etc. Makes the code easier to read, less like a complex math formula :)

lucifer1004: The usage of Map<K, V> and Set<U> is consistent with other Deep-* types.

But I think I should rename P to Filter. This change can help a lot.

lucifer1004: @quezak I've made a new commit, please see if it is better now.

README.md
```diff
@@ -27,11 +27,14 @@ yarn add --dev ts-essentials
 
 - [Install](#Install)
 - [What's inside?](#Whats-inside)
-  - [Basic:](#Basic)
+  - [Basic](#Basic)
```
krzkaczor: <3

lucifer1004:
One more thing to explain, I changed the markup type (for properties requiring complete deletion) from `true` to `never` in an earlier commit, because I found that when I used `true` as the markup type, `never` would just fit in as well. It seemed that the easiest way to address this was to change the markup type to `never`, and I did so.

quezak:
> I changed the markup type from `true` to `never`

I noticed that, nice idea -- gives a better feeling when writing the filter!

krzkaczor:
Thank you for this contribution 🙏 Released as `ts-essentials@3.0.0-beta.1`. 

Btw. I am gonna update readme `master`'s readme about what's cooking here.

krzkaczor:
@all-contributors please add @lucifer1004 for documentation

krzkaczor:
@all-contributors please add @quezak for documentation, review

PR #74, opened 2019-07-04
Add Function as a check before object to prevent #73

### LINKED ISSUE #73: DeepRequire on type with function on a property sets that property to {}
Below is an example of what I tried to do and the results received. I expect the final notation to be what happens instead.

```ts
type PartialObject = {
  a?: string
  b?: string
  c?: (input: string) => void
}

type RequiredObject = DeepRequired<PartialObject>

// actual type output
type RequiredObject = {
  a: string;
  b: string;
  c: {};
}

// expected type output
type RequiredObject = {
  a: string;
  b: string;
  c: (input: string) => void;
}

kevinpeno:
fixes #73

### DIFF
--- a/lib/types.ts
+++ b/lib/types.ts
@@ -25,6 +25,8 @@ export type DeepRequired<T> = T extends Primitive
   ? NonNullable<T>
   : T extends any[]
   ? DeepRequiredArray<NonNullable<T[number]>>
+  : T extends Function
+  ? T
   : T extends {}
   ? { [K in keyof T]-?: DeepRequired<NonNullable<T[K]>> }
   : T;


krzkaczor:
Released as `ts-essentials@2.0.12`

krzkaczor:
Btw. @kevinpeno are u using typescript@3.5? We are wondering about making 3.5 a requirement for `ts-essentials@3` and we are curious how the community reacts to it. You can chip in https://github.com/krzkaczor/ts-essentials/issues/70

krzkaczor:
@all-contributors please add @kevinpeno for code

PR #81, opened 2019-07-15
Use internal _Omit to keep it compatible with TS < 3.5

xaviergonz:
I have a project that uses TS 3.5, but sadly typedoc (the package to generate documentation from sources) is stuck with TS 3.2.4.

Since the project has to compile with both I'm stuck, since:
* If I don't declare a global Omit to polyfill the type then the TS3.5 compilation will be ok, but the 3.2 one will complain that the type def of this package is missing Omit
* If I do declare a global Omit then the TS3.2 compilation will be ok, but the 3.5 one will complain that Omit is already declared

This PR fixes this by using an internal _Omit type which is a copy of the one from the typdefs of TS 3.5

### DIFF
--- a/lib/types.ts
+++ b/lib/types.ts
@@ -1,6 +1,13 @@
 /** Essentials */
 export type Primitive = string | number | boolean | bigint | symbol | undefined | null;
 
+/**
+ * Construct a type with the properties of T except for those in type K.
+ * Use this instead of TS 3.5 Omit to keep it backwards compatible.
+ */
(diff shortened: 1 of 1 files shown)


quezak:
If you just want to make it work with older TS versions, you can use `ts-essentials@2`. We will be maintaining the `2.x` branch for a few months just for cases like this. Unless there are features in the `3.x` release that you urgently need together with old TS support?

I checked [typedoc's deps](https://github.com/TypeStrong/typedoc/blob/master/package.json), and they already require TS 3.5.x on their `0.15.x` release, available on the `@next` channel -- maybe you can just update typedoc? :)

Issue #85, opened 2019-08-08
Accept anything except empty object


EduardoRFS:
I was thinking about a type that accepts anything, except empty objects, something like that, it will be really useful for generic functions:

```typescript
type X<T extends {}> = keyof T extends never ? never : T;
// usage
const fn = <T>(t: X<T>): T => t;

fn({}); // fail
fn({ a: 1 }); // pass
```

But I don't know a good name, I was thinking about `NonEmptyObject` but it seems silly.

krzkaczor:
I think that this tslint rule can be related to this idea: https://palantir.github.io/tslint/rules/no-inferred-empty-object-type/ 

Does it help in your use case?

EduardoRFS:
Nope because it is for an API, after filtering props using PickProperties it can result in an empty object, but an empty object is useless for the API so refusing it is a way to say "this object doesn't match the constrains".

I feel like it would be useful for a bunch of different APIs essentially anything that iterate over keys and is expecting a object with keys.

krzkaczor:
@EduardoRFS could you present please an example of code where this would be useful? Just to make sure we are on the same page.

@quezak thoughts?

EduardoRFS:
[example](https://github.com/EduardoRFS/mali/blob/mali-t/index.d.ts)

Isn't a great example, but after you try to use it, gosh it is fun, you just pass the entire file and he can verify it the file is what is expected and get only the signatures that is reasonable, also typings of context so you know what you should read and write.

Also a thing that can be useful is the GetOwnProperties, is a way to get only the properties of a child not his parent

BTW that's why I did the PR for PickProperties

quezak:
I see some use cases for this, where you need to provide _at least one_ of some optional properties in a generic context, so I'd accept a PR for that :)

As for the name, `NonEmptyObject` sounds OK -- personally I use `Nonempty` as a single word, but a quick google search proves that `non-empty` is probably more popular.

EduardoRFS:
But I was thinking there is a thing, what is the intent of this project? Because it isn't strictly "essential", or it will be a minimal alternative to `ts-toolbelt`? One advantage for me is that ts-toolbelt create terrible error messages and this project creates only bad error messages

krzkaczor:
@EduardoRFS `ts-toolbelt` or `type-fest` was created later so I am not sure how they want to differentiate. 

[...]

PR #86, opened 2019-08-14
implement NonEmptyObject

EduardoRFS:
Related to #85

### DIFF
--- a/README.md
+++ b/README.md
@@ -312,6 +312,21 @@ type GetDefined<TypesMap extends { [key: string]: any }> = keyof NonNever<
 >;
 ```
 
+### NonEmptyObject
+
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -312,6 +312,21 @@ type GetDefined<TypesMap extends { [key: string]: any }> = keyof NonNever<
 >;
 ```
 
+### NonEmptyObject
+
+Useful for accepting only objects with keys, great after a filter like OmitProperties or PickProperties.
+
+```typescript
+/* return never if the object don't have any number value*/
```
quezak: (grammar-nazi-nitpick) `don't` -> `doesn't`

EduardoRFS: fixed <3

krzkaczor:
Thanks for this! Published as `ts-essentials@3.0.2`

PR #94, opened 2019-10-17
Add Buildable

a1russell:


a1russell:
> Code looks OK.
> 
> @a1russell what are the arguments for using this instead of just `DeepPartial<DeepWritable<T>>`? (is it just a convenience shorthand, or are there some other benefits?)

I figured it would be a bit more efficient to go through a nested structure once rather than twice, but they should be pretty much the same, yeah.

.gitignore
```diff
@@ -1,2 +1,8 @@
-node_modules
-dist
+.idea/
+dist/
+node_modules/
+*.iml
+*.ipr
+*.iws
+*.sw?
+.DS_Store
```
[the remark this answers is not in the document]
a1russell: Ok, sure, I'll remove this commit.

krzkaczor:
Few thoughts:
1. 

>I figured it would be a bit more efficient to go through a nested structure once rather than twice, but they should be pretty much the same, yeah.

I definitely would advise against optimizations like this (no benchmarks, no perf issues etc).
It would be much cleaner to just implement it as `DeepPartial<DeepWritable<T>>`. 

2. What are the good usecases for such type? (this should be part of the docs). I think this type can prove it's usefulness with things like configs that have defaults that later are deep merged together. Anything else?

3. `Buildable` is such a generic name :/

a1russell:
1. We're talking about the difference of 10 lines of code. I don't think this is that much of an optimization or complex addition. If it were implemented as `DeepPartial<DeepWritable<T>>` I wouldn't even bother with the PR.
2. Personally I use Buildable more than either DeepWritable or DeepPartial, which is why I thought this might be worth contributing. You can make up your own mind whether it's worth accepting. I'll use it in my own code anyway.
3. Generic name for a generic type. Open to suggestions, though.

Not going to hurt my feelings if you reject the PR. I was just trying to contribute.

krzkaczor:
Thanks for this idea! I think that we need a little bit more community feedback for this one, before merging. Personally, I never had a need for such type (but I am not saying that it's not useful ;) ).

1. Find more good usecases and explain them in docs
2. Perhaps figure out a better name based on those usecases.

Do you have by any chance any open source code that uses this type?

Thanks again for your input!

yxliang01:
I find this type legit and the name quite conforms with the definition. There are many cases where building an object step by step is more elegant then assigning the values into many intermediate variables then compose the final object at once.

a1russell:
My main use case is as @yxliang01 has intuited. To build an object step by step. For example:

```typescript
interface ReadonlyObject extends Readonly<{
  simple: number;
  nested: Readonly<{
    a: string;
    array: ReadonlyArray<Readonly<{ bar: number }>>;
  }>;
}> {}
const buildable: Buildable<ReadonlyObject> = {};
buildable.nested = {};
buildable.nested.a = 'test';
buildable.nested.array = [];
buildable.nested.array.push({ bar: 1 });
buildable.simple = 7;
const finished = buildable as ReadonlyObject;
```

The name is due to this use case and its similarity to the builder pattern.

As for whether I have any open source code that uses this, the answer is no. I primarily work on proprietary software, and this is used in lots of business logic there.

README.md
```diff
@@ -160,6 +161,23 @@ test[0].foo = "b";
 test[0].bar.x = 2;
 ```
 
+### Buildable
+
+A combination of both `DeepWritable` and `DeepPartial`.
+
```
yxliang01:
```suggestion
This type allows building an object step by step by assigning the values to its attributes in multiple statements.
```

a1russell: Shall I update the example as well to something more similar to above?

yxliang01: Indeed. It will look better. Maybe a comparison between two :)

krzkaczor:
Now I get it, I missed a clear explanation where this type can be useful. 

@a1russell can you add more explanation to readme (what @yxliang01 suggested?) and we can merge it.

Thanks!

krzkaczor:
@all-contributors please add @a1russell for code,docs

krzkaczor:
@all-contributors please add @a1russell for code, docs

krzkaczor:
Thanks @a1russell! 🎉 

Released as `ts-essentials@3.0.3`

Issue #99, opened 2019-11-05
Simplify type definitions via TS 3.7 recursive type aliases


lucifer1004:
See [here](http://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html?ref=hvper.com#more-recursive-type-aliases).

By leveraging the recursive type feature introduced in TS 3.7, we no longer need intermediate interfaces to express recursive structures, which can greatly simplify our code base.

Maybe we can open an experimental branch for this, since this would definitely be a breaking change (we will rely on TS 3.7).

quezak:
Yes, definitely!

lucifer1004:
Addressed via #105

PR #105, opened 2019-12-04
refactor: simplify Deep-* types via TS 3.7 recursive types (#99)

lucifer1004:
All Deep-* types have been greatly simplified, including `DeepOmit`.

This PR requires TypeScript 3.7, which is a breaking change.

### DIFF
--- a/lib/types.ts
+++ b/lib/types.ts
@@ -9,53 +9,45 @@ export type DictionaryValues<T> = T extends Dictionary<infer U> ? U : never;
 export type DeepPartial<T> = T extends Primitive | Function | Date
   ? T
   : T extends Map<infer K, infer V>
-  ? DeepPartialMap<K, V>
+  ? Map<DeepPartial<K>, DeepPartial<V>>
   : T extends Set<infer U>
-  ? DeepPartialSet<U>
+  ? Set<DeepPartial<U>>
   : T extends {}
   ? { [K in keyof T]?: DeepPartial<T[K]> }
   : Partial<T>;
-interface DeepPartialSet<ItemType> extends Set<DeepPartial<ItemType>> {}
-interface DeepPartialMap<KeyType, ValueType> extends Map<DeepPartial<KeyType>, DeepPartial<ValueType>> {}
 
 /** Like NonNullable but recursive */
 export type DeepNonNullable<T> = T extends Primitive | Function | Date
   ? NonNullable<T>
   : T extends Map<infer K, infer V>
-  ? NonNullableMap<K, V>
+  ? Map<DeepNonNullable<K>, DeepNonNullable<V>>
   : T extends Set<infer U>
-  ? NonNullableSet<U>
+  ? Set<DeepNonNullable<U>>
   : T extends {}
   ? { [K in keyof T]: DeepNonNullable<T[K]> }
   : NonNullable<T>;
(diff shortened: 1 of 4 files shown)


lib/types.ts
```diff
@@ -138,9 +124,7 @@ export type MarkRequired<T, RK extends keyof T> = Exclude<T, RK> & Required<Pick
 export type MarkOptional<T, K extends keyof T> = Omit<T, K> & Partial<Pick<T, K>>;
 
 /** Convert union type to intersection #darkmagic */
-export type UnionToIntersection<U> = (U extends any ? (k: U) => void : never) extends ((k: infer I) => void)
-  ? I
-  : never;
+export type UnionToIntersection<U> = (U extends any ? (k: U) => void : never) extends (k: infer I) => void ? I : never;
```
quezak: pls undo this formatting change, even thought the brackets and newlines are not explicitly necessary,  they did considerably improve readability of this complex expression

lib/types.ts
```diff
@@ -158,7 +142,7 @@ export type AsyncOrSync<T> = PromiseLike<T> | T;
 // This potentially abuses compiler some inconsistencies in checking type equality for generics,
 // because normally `readonly` doesn't affect whether types are assignable.
 // @see https://stackoverflow.com/a/52473108/1815209 with comments
-type IsEqualConsideringWritability<X, Y> = (<T>() => T extends X ? 1 : 2) extends (<T>() => T extends Y ? 1 : 2)
+type IsEqualConsideringWritability<X, Y> = (<T>() => T extends X ? 1 : 2) extends <T>() => T extends Y ? 1 : 2
```
quezak: same here

quezak:
Looks good besides the one formatting comment, good timing since I wanted to hop on to this myself :)

@krzkaczor are we going to do a major release straight away, or merge this to a branch and release as `@next` first?

lib/types.ts
krzkaczor: I think this is breaking change in prettier so I don't think that we can do anything about it...

package.json
```diff
@@ -26,9 +26,9 @@
...
   "devDependencies": {
     "@codechecks/build-size-watcher": "^0.1.0",
-    "@codechecks/client": "^0.1.4",
-    "conditional-type-checks": "^1.0.1",
-    "prettier": "^1.18.2",
-    "typescript": "^3.5.2"
+    "@codechecks/client": "^0.1.10",
+    "conditional-type-checks": "^1.0.4",
+    "prettier": "^1.19.1",
+    "typescript": "^3.7.2"
```
quezak: one more thing: you need to also update TS version in the `peerDependencies` above

krzkaczor:
@quezak I'll do a major release right after its merged.

lib/types.ts
lucifer1004: Yes, the format is done by prettier.

package.json
lucifer1004: Done.

lib/types.ts
quezak: :( that's why I "really don't like" it, but no choice here then

krzkaczor: > OK BOOMER 💥 

😉

krzkaczor:
Nice one, thanks!

PR #108, opened 2019-12-09
Improve docs by adding alliasses for some types

krzkaczor:
Closes https://github.com/krzkaczor/ts-essentials/issues/102

### DIFF
--- a/README.md
+++ b/README.md
@@ -164,6 +164,8 @@ test[0].bar.x = 2;
 
 ### Buildable
 
+*keywords: builder*
+
 A combination of both `DeepWritable` and `DeepPartial`.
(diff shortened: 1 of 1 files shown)


quezak:
A few more ideas, feel free to include what you feel fits:
* Dictionary: *map, mapping*
* Deep Partial & others: *nested, optional*
* Omit: *exclude, skip* (not sure about the first one -- it's a good keyword, but `Exclude` in TS has a different meaning so it could be confusing)
* Deep Omit: *omit nested*
* OmitProperties: same as Omit
* Merge: *override* [?]
* MarkOptional: *mark nullable* [?]

quezak:
CC @macbem @akwodkiewicz

PR #109, opened 2019-12-09
Add assertion functions

krzkaczor:
Closes https://github.com/krzkaczor/ts-essentials/issues/106

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,6 +47,7 @@ npm install --save-dev ts-essentials
   - [Exhaustive switch cases](#Exhaustive-switch-cases)
   - [ValueOf type](#ValueOf-type)
   - [AsyncOrSync type](#AsyncOrSync-type)
+  - [Assertions](#Assertions)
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -559,6 +560,24 @@ class Travis implements CiProvider {
...
+Simple runtime assertion that narrows involved types using [assertion functions](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#assertion-functions).
+
+Note: This function is not purely type level and leaves minimal runtime trace in generated code.
+
+```typescript
+const something: string | undefined = "abc" as any
+assert(something, "Something has to be defined!")
+// from now on `something` is string, if this wouldn't be a case, assert would throw
+
+const anything: any = "abc" as any
```
quezak: isn't `...: any = ... as any` redundant? (I mean, specifying `any` once should be sufficient)

Issue #113, opened 2019-12-13
Add `Error` and `RegExp` as built-in types


lucifer1004:
Currently, we have `Primitive`, `Date` and `Function`.

`Error` and `RegExp` can also be added.

Issue #111, opened 2019-12-13
WeakMap/WeakSet/Promise support in Deep-* types


lucifer1004:
Currently, Deep-* types only take into account `Map` and `Set`, so when applied to types containing `WeakMap` or `WeakSet`, an error will occur.

lucifer1004:
Support for `Promise<T>` can also be added.

PR #114, opened 2019-12-13
feat: Deep-* support for WeakMap/WeakSet/Promise/Error/RegExp

lucifer1004:
See #111 #113

### DIFF
--- a/lib/types.ts
+++ b/lib/types.ts
@@ -1,50 +1,74 @@
 /** Essentials */
-export type Primitive = string | number | boolean | bigint | symbol | undefined | null;
+export type Builtin = string | number | boolean | bigint | symbol | undefined | null | Function | Date | Error | RegExp;
 
 /** Dictionaries related */
 export type Dictionary<T, K extends string | number = string> = { [key in K]: T };
 export type DictionaryValues<T> = T extends Dictionary<infer U> ? U : never;
 
 /** Like Partial but recursive */
-export type DeepPartial<T> = T extends Primitive | Function | Date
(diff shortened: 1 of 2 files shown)


quezak:
Looks good -- though I don't like the fact that the `Deep*` types have over 9000 cases each, it's probably inevitable if we want them to be truly universal.

quezak:
One question though: is treating `Error` as a builtin type a good idea? One could have a custom error with custom object properties, and still want to have a `DeepPartial` of that.

lucifer1004:
> Looks good -- though I don't like the fact that the `Deep*` types have over 9000 cases each, it's probably inevitable if we want them to be truly universal.

Yes... We really need a generic Deep type, but it still seems impossible right now, unless we have something like generic-generic types.

lucifer1004:
> One question though: is treating `Error` as a builtin type a good idea? One could have a custom error with custom object properties, and still want to have a `DeepPartial` of that.

Actually similar things can also happen to other types. Users might extend a JS type and add some custom properties, but we cannot distinguish the modified versions from the raw ones using something like <T extends Error>.

quezak:
> Actually similar things can also happen to other types. Users might extend a JS type and add some custom properties, but we cannot distinguish the modified versions from the raw ones using something like .

Yes, but I see no rational cases to have a partial of a `Function`, `RegExp` or `Date`, and surely all the primitives, while for `Error` it might actually make sense. @krzkaczor what do you think?

krzkaczor:
I don't think that `Error` subtyping edge case is something we should worry about now. Thanks for your contribution!

Issue #112, opened 2019-12-13
What to be expected from DeepPartial<Array<T>>?


lucifer1004:
```typescript
type Student = Array<{name: string, age: number}>
type PartialStudent = Partial<Student> 
// this yields ({
//    name: string;
//    age: number;
//} | undefined)[]
type DeepPartialStudent = DeepPartial<Student>
// this yields ({
//    name?: string;
//    age?: number;
//} | undefined)[]
```

Is this to be expected? Applying `Partial` or `DeepPartial` on an `Array` allows array elements to be `undefined`? @krzkaczor @quezak

lucifer1004:
The advantage of current implementation (not handling `Array` as a special case) is that `tuple` can be handled neatly. 
If `Array` is handled like `Map` or `Set`, Deep-* on a tuple will not work as expected.

quezak:
The intuitive way for me would be that `DeepPartial<Array<T>>` gives you just `Array<DeepPartial<T>>`.

Let's see what others have to say about tuples. Personally for me, if you need a deep partial on a tuple with different element types, it's probably a sign of bad code design, and the object should be an actual object with named keys instead.

/summon @macbem

krzkaczor:
I also find that DeepPartial<array> is gonna produce array of undefines at least weird. 

I would be happy to change this behavior and release as breaking change OR maybe introducing a new type? Also, should be there a consistency with other `Deep-*` types?

lucifer1004:
If we are to change this behavior, I think all Deep-* types should be changed.

krzkaczor:
Released as `ts-essentials@5.0.0`

PR #116, opened 2019-12-18
[Discussion] fix: DeepPartial on Array (#112)

lucifer1004:
I only implemented the change on `DeepPartial`, because current implementation of other `Deep-*` types have no problem dealing with `Array`, while the change will ruin `Tuple`.

I have also changed the related tests. @krzkaczor @quezak

### DIFF
--- a/lib/types.ts
+++ b/lib/types.ts
@@ -1,9 +1,22 @@
 /** Essentials */
 export type Primitive = string | number | boolean | bigint | symbol | undefined | null;
 export type Builtin = Primitive | Function | Date | Error | RegExp;
+export type IsTuple<T> = T extends [infer A]
+  ? T
+  : T extends [infer A, infer B]
+  ? T
+  : T extends [infer A, infer B, infer C]
+  ? T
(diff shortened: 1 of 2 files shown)


krzkaczor:
I am not sure what to do about this one... I don't use tuples too often so I would prefer to have proper support for arrays first, but breaking tuples also doesn't seem like a great idea... 

General question: is there a way to find if T is a tuple but not array? Maybe something like `T extends [infer X] ... T extends [infer X, infer Y]` etc. should work. We could support up to 5 parameters and otherwise treat it as array. WDYT?

lucifer1004:
> I am not sure what to do about this one... I don't use tuples too often so I would prefer to have proper support for arrays first, but breaking tuples also doesn't seem like a great idea...
> 
> General question: is there a way to find if T is a tuple but not array? Maybe something like `T extends [infer X] ... T extends [infer X, infer Y]` etc. should work. We could support up to 5 parameters and otherwise treat it as array. WDYT?

I have made a new commit, which treats tuples with 5 or fewer parameters separately.

PR #121, opened 2020-01-16
add XOR type

DevilZh:
Gets the XOR(Exclusive-OR) type which could make 2 types exclude each other
```typescript
  type A = {a: string}
  type B = {a: number; b: boolean}
  type C = {c: number}

  let A_XOR_B: XOR<A, B>
  let A_XOR_C: XOR<A, C>

  // fail
  A_XOR_B = {a: 0}
  A_XOR_B = {b: true}
  A_XOR_B = {a: '', b: true}
  A_XOR_C = {a: '', c: 0}

  // ok
  A_XOR_B = {a: 0, b: true}
  A_XOR_B = {a: ''}
  A_XOR_C = {c: 0}
```

### DIFF
--- a/lib/types.ts
+++ b/lib/types.ts
@@ -224,3 +224,9 @@ export type ReadonlyKeys<T extends object> = {
 export type WritableKeys<T extends {}> = {
   [P in keyof T]-?: IsFullyWritable<Pick<T, P>> extends true ? P : never;
 }[keyof T];
+
+/** Mark some properties which only the former including as optional and set the value to never */
+type Without<T, U> = { [P in Exclude<keyof T, keyof U>]?: never };
+
+/** get the XOR type which could make 2 types exclude each other */
+export type XOR<T, U> = (T | U) extends object ? (Without<T, U> & U) | (Without<U, T> & T) : T | U;
--- a/lib/types.ts
+++ b/lib/types.ts
@@ -229,4 +229,4 @@ export type WritableKeys<T extends {}> = {
 type Without<T, U> = { [P in Exclude<keyof T, keyof U>]?: never };
 
 /** get the XOR type which could make 2 types exclude each other */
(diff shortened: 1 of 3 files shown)


lib/types.ts
```diff
@@ -224,3 +224,9 @@ export type ReadonlyKeys<T extends object> = {
 export type WritableKeys<T extends {}> = {
   [P in keyof T]-?: IsFullyWritable<Pick<T, P>> extends true ? P : never;
 }[keyof T];
+
+/** Mark some properties which only the former including as optional and set the value to never */
+type Without<T, U> = { [P in Exclude<keyof T, keyof U>]?: never };
```
quezak: The `Without` name of this helper type is, at least for me, not descriptive at all :(

quezak:
I must honestly say, I'm surprised. My first thought were "but wait, the `fail` cases should already fail with just `A | B`" -- while cases 3 and 4 are allowed :astonished:

While it's obvious that until you determine if it's an `A` or a `B`, you can only use `A_XOR_B.a` and it will be of type `string | number`, and you can't use `A_XOR_B.b`, since you don't know if it's even there, I expected that you can't assign such a "hybrid typed" value.

lib/types.ts
DevilZh: > The `Without` name of this helper type is, at least for me, not descriptive at all :(

so, what do u think? PickOverlapToNever? ... I think it‘s too redundant.

quezak: If it's not _too_ long, it's better to have a kinda redundant name that helps to understand the type, than a short one that doesn't. But "overlap" is also not accurate, since you're _excluding_ the overlap it would rather be "pick excluded to never", but that starts to sound weird :D  You can also check out my reimplementation -- it may be that this helper type is too compilicated to begin with. But maybe I'm overreacting here -- @krzkaczor @macbem what are your opinions?

krzkaczor:
Good work guys! XOR seems to be useful.

krzkaczor:
Released as `ts-essentials@5.0.0`

PR #122, opened 2020-01-27
Add support for ReadonlyMap and ReadonlySet

mikhailswift:


lib/types.ts
```diff
@@ -97,10 +113,14 @@ export type DeepWritable<T> = T extends Builtin
   ? T
   : T extends Map<infer K, infer V>
   ? Map<DeepWritable<K>, DeepWritable<V>>
+  : T extends ReadonlyMap<infer K, infer V>
```
krzkaczor: Shouldn't it become just `Map` for writable types? These `Readonly*` types require special care IMO...

@quezak WDYT?

quezak: yeah, I think it should -- writable is writable

quezak: but, if we do this, `DeepReadonly` for a regular `Map` should also return `ReadonlyMap<K, DeepReadonly<V>>`, for consistency

mikhailswift: That all makes sense.  Made those changes.

PR #125, opened 2020-01-30
feat: DeepNullable

### LINKED ISSUE #124: Feature request: DeepNullable
What about implementing a `DeepNullable` type such that, given a `MyType` type:

```
interface MyType {
  field1: string;
  field2: string;
  field3: {
    field4: string;
    field5: number;
  };
}
```

the `DeepNullable<MyType>` would be equal to:

```
  field1: string | null;
  field2: string | null;
  field3: {
    field4: string | null;
    field5: number | null;
  };
```

this is different than `DeepPartial` as properties should never be undefined.

EDIT: also non-primitive properties should always be defined. So when you navigate through the object you should never get "unable to access XXX of undefined".

FrancescoBorzi:
Help wanted: for some reasons I can't get the test working and I'm not sure what the problem is

- Closes #124


quezak:
@FrancescoBorzi just look at the "[Details](https://circleci.com/gh/krzkaczor/ts-essentials/273?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link)" link next to the CI build. The test aren't even running because the formatting check failed. Did you check the instructions at [CONTRIBUTING.md](https://github.com/krzkaczor/ts-essentials/blob/master/CONTRIBUTING.md)?

FrancescoBorzi:
@quezak yes but this is the problem I'm facing:

image

and I don't know what is causing it, that's all I get:

image

so I don't know what exactly is wrong

quezak:
Ok, I worked my way up from smaller examples and found two things:

1. In your implementation for the `{}` variant (a deep nullable object) uses optional fields, while they should be required, but with a nullable value type:

```typescript
/** Recursive nullable */
export type DeepNullable<T> = T extends Builtin
  ? T | null
  : T extends Map<infer K, infer V>
  ? Map<DeepNullable<K>, DeepNullable<V>>
  : T extends WeakMap<infer K, infer V>
  ? WeakMap<DeepNullable<K>, DeepNullable<V>>
  : T extends Set<infer U>
  ? Set<DeepNullable<U>>
  : T extends WeakSet<infer U>
  ? WeakSet<DeepNullable<U>>
  : T extends Array<infer U>
  ? T extends IsTuple<T>
    ? { [K in keyof T]: DeepNullable<T[K]> | null }  // removed `?` on key, but added `| null` on value
    : Array<DeepNullable<U>>
  : T extends Promise<infer U>
  ? Promise<DeepNullable<U>>
  : T extends {}
  ? { [K in keyof T]: DeepNullable<T[K]> | null }  // same here
  : T | null;
```

2. The test nested example you provided had actually _too many_ nulls. For example, in `array`, only the element keys should be nullable, not the key itself. It makes sense to me -- say you have a `user.posts` field. In `DeepNullable<User>`, `posts` should be either `null` or an array of posts with nullable properties -- `null` can't be an element of that array.  Same change goes for set keys, map values, etc. Working example:
```typescript
type ComplexNestedNullable = {
  simple: number | null;
  nested: {
    date: Date | null;
    func: (() => string) | null;
    array: { bar: number | null }[] | null;
    tuple: [string | null, number | null, { good: boolean | null } | null] | null;
    set: Set<{ name: string | null }> | null;
    map: Map<string | null, { name: string | null }> | null;
    promise: Promise<{ foo: string | null; bar: number | null }> | null;
  } | null;
};
```

Please check if that implementation works with your usecases (you must have some, since you proposed the feature, right? :smiley: ), fix formatting and add a documentation entry with a small example.

FrancescoBorzi:
@quezak thanks a lot for your help, I appreciate!

I've updated the PR with your suggestions as well as adding a test for my use case, which is simpler than the complex object used for other test cases.

Ready to be reviewed!

quezak:
@FrancescoBorzi Thanks for the update! If you add a brief description to the README, looks like we'll be ready to go :muscle:

FrancescoBorzi:
@quezak done, thanks!

README.md
```diff
@@ -134,6 +134,42 @@ const sampleNonNullable: ComplexObjectNonNullable = {
     array: [{bar: null}], // Error: Type 'null' is not assignable to type 'number'
   }
 }
+
+type SimpleType = {
```
krzkaczor: please update title of this readme section to add info about this type

FrancescoBorzi: @krzkaczor done

quezak:
Updated the docs heading a bit, looks ready now, thanks @FrancescoBorzi!

@all-contributors please add @FrancescoBorzi for code, docs

FrancescoBorzi:
thanks a lot for your help @quezak @krzkaczor

krzkaczor:
Released as `v6.0.1` 👍

PR #131, opened 2020-02-07
chore: update .gitignore

FrancescoBorzi:
- `package-lock.json` is not used (`yarn.lock` is used instead) so imho we should ignore it
- `.idea` is automatically generated by all IntelliJ-based IDEs (e.g. Webstorm)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,4 @@
 node_modules
 dist
+package-lock.json
+.idea


.gitignore
```diff
@@ -1,2 +1,4 @@
 node_modules
 dist
+package-lock.json
```
quezak: this project uses `yarn`, not `npm`, so why would this file ever be created? @krzkaczor what do you think?

.gitignore
```diff
@@ -1,2 +1,4 @@
...
+.idea
```
quezak: I've personally received a nope on this (for vim/vscode though) form @krzkaczor once before 😅 but his arguments are right: you can add those to your [global gitignore](https://stackoverflow.com/questions/7335420/global-git-ignore) -- otherwise soon we'll end up with dozens of gitignore entries for every OS & editor combination out there.

krzkaczor:
Sorry, can't do ;) 

We don't use npm so we don't care about `package-lock.json` (you should simply use yarn). And files regarding personal preferences (IDEs, OS realated files) should be part of your global git ignore. Otherwise, we will end up with a list of most popular IDEs directly in our gitignore file ;)

PR #133, opened 2020-02-28
allow SafeDictionary<T, number>

quezak:
`SafeDictionary` is a really great recent addition, but I found out that omitting the `K` param is actually insufficient -- I'm using `Dictionary<T, number>` in multiple places, and I'd like to switch them too to the safe variant.

CC @sz-piotr since you've introduced this.


krzkaczor:
Released as `ts-essentials@6.0.2`

PR #136, opened 2020-03-25
Simplify Merge type

kubk:
Merge type could be simplified. This is the definition of TypeScript's built-in Extract type: 
```
type Extract<T, U> = T extends U ? T : never;
```
That's why these types are equal
```
type Test = Extract<'id' | 'name' | 'age', 'id' | 'name'>; // Equals to 'id' | 'name'
type Test2 = 'id' | 'name' // Equals 'id' | 'name'
```


quezak:
I have a very vague memory that I wanted to propose this last year, and finally found a counterexample. I'll take a detailed look tomorrow and see if I can recreate it or if I'm mistaken after all :)

krzkaczor:
@all-contributors please add @kubk for code

krzkaczor:
Thanks @kubk! Released as 📦 `ts-essentials@6.0.4`

Issue #135, opened 2020-03-24
Suggestion: add `noop` function


leaumar:
```typescript
function noop(...args: unknown[]): void {
    void args;
}

// Result:
// const x = noop(0, "", false, {}, []); // undefined
```

Unlike the existing `noop-ts`, this function returns `void`. It also avoids the use of `any` altogether, and does not have unused parameters this way (for this project's linting).

When is it handy:
```typescript
// reassign to real callback later - avoids null placeholder
let callback: (event) => void = noop;
// idem
const callback: (event) => void = mapOfCallbacksForEvents.get("myEvent") ?? noop;

function doSomething(): Promise<void> {
    // "delete" T, return Promise<void>
    return getSomething<T>().then(noop);
    // for when the alternative is
    // `await getSomething(); break;`
    // or
    // `await getSomething(); return undefined;`
    // due to code style
}

try {
   dontCareIfThisFails();
} catch(mootError) {
    // linting does not complain about unused variable, and readers get a clear hint of intention
    noop(mootError);
}
```

quezak:
While a `noop` function can be useful, it seems to me like you're using it for wrong reasons -- as a workaround for linter/style problems that shouldn't be there in the first place or have clearer solutions :smiley: 

1. Using `void args` to silence an unused args warning looks like a dirty hack, also it's a _runtime_ workaround for a _non-runtime_ problem. Linter should be disabled for this line instead -- that's what the `eslint-disable-next-line` comments are for. Alternatively, I'm using a [linter config](https://eslint.org/docs/rules/no-unused-vars#argsignorepattern) that allows arg names prefixed with `_` to be unused. And many projects [actually forbid](https://eslint.org/docs/rules/no-void) using the `void` operator anyway.

2. If I saw `return asyncFn().then(noop)` in a code review I would be confused and ask the author what's the intention (being suspicious of some under-the-hood magic required there). This can be done in a more readable way if you need this pattern frequently, but the exact solution will depend on the author's personal preferences. For example, `return ignoreResult(asyncFn())`, `return asyncFn().then(ignoreResult)`, even `return void (await asyncFn())` (though it looks very hacky and I'm not a fan of the `void` operator), or just make the outer function return `Promise<any>` and just return without any "runtime casting".

[...]

PR #137, opened 2020-03-25
#135: add noop

leaumar:
implement #135

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,6 +58,7 @@ npm install --save-dev ts-essentials
 ### Basic
 
 - `Primitive` type matching all primitive values.
+- `noop` function that takes any arguments and returns nothing, as a placeholder for e.g. callbacks.
 
 ### Dictionaries
 
--- a/test/index.ts
+++ b/test/index.ts
@@ -20,6 +20,7 @@ import {
   Merge,
   NonEmptyObject,
   NonNever,
+  noop,
   PickProperties,
   ReadonlyKeys,
   SafeDictionary,
@@ -446,4 +447,11 @@ function testXOR() {
 
   type Test1 = Assert<IsExact<Actual1, Expected1>>;
   type Test2 = Assert<IsExact<Actual2, Expected2>>;
+}
+
+function testNoop() {
+  const x: void = noop();
+  const y: void = noop(false, 0, "", {}, [], null, undefined, Promise.resolve(), new Error(), noop);
+  const z: Promise<void> = Promise.resolve("foo").then(noop);
+  const callback: (_: number) => void = noop;
 }
--- a/lib/functions.ts
+++ b/lib/functions.ts
@@ -9,3 +9,5 @@ export function assert(condition: any, msg: string = "no additional info provide
     throw new Error("Assertion Error: " + msg);
   }
 }
+
+export function noop(..._args: unknown[]): void {}


test/index.ts
```diff
@@ -20,6 +20,8 @@ import {
   Merge,
   NonEmptyObject,
   NonNever,
+  noop,
+  noop as erase,
```
macbem: This line is unnecessary - you can just use the previous import.

quezak: yes, also the name `erase` is misleading, as it suggests that the function _does something_

test/index.ts
```diff
@@ -447,3 +449,10 @@ function testXOR() {
   type Test1 = Assert<IsExact<Actual1, Expected1>>;
   type Test2 = Assert<IsExact<Actual2, Expected2>>;
 }
+
+function testNoop() {
+  const x: void = noop();
+  const y: void = noop(false, 0, "", {}, [], null, undefined, Promise.resolve(), new Error(), noop);
+  const z: Promise<void> = Promise.resolve("foo").then(erase);
+  const callback: (_: number) => void = Math.random() < 0.5 ? (i => void i) : noop;
```
quezak: the `void` operator is weird ;)
```suggestion
  const callback: (_: number) => void = Math.random() < 0.5 ? (() => {}) : noop;
```

test/index.ts
leaumar: I guess this file isn't used as any form of user documentation, so providing a semantic example use isn't meaningful, you're right

test/index.ts
leaumar: I always find `{}` dubious since it looks like an empty code block or an object literal at the same time, but sure

leaumar: Also, I guess if this file isn't meant to serve as documentation, the `i` callback for assignment equality showcasing isn't useful

leaumar:
I did run `test:fix` several times, but then that one last little edit I did without running it got me :)

README.md
```diff
@@ -58,6 +58,7 @@ npm install --save-dev ts-essentials
 ### Basic
 
 - `Primitive` type matching all primitive values.
+- `noop` function that takes any arguments and returns nothing. Makes a convenient default callback, Promise value eraser, etc.
```
leaumar: should this wording perhaps avoid the use of "etc" and just be 2 examples without leading on?

macbem: TBH I'd avoid listing `Promise value eraser` as an example - you're the first person I've seen using this naming. A `noop` is a `noop` - a function that accepts anything, does nothing and returns nothing, just a convenient placeholder fn.

krzkaczor:
Hey, sorry to jump on this so late, after the main discussion. Two things:

1. Have we discussed returning `any` instead of `void`? This way it could be used as a dummy implementation for anything (a littel bit different than simple noop).

I thought that we could combine this with https://github.com/krzkaczor/ts-essentials/issues/134 

Basically we could have: 
```
export const AnyValue = undefined as any;
export const AnyFunc = (): any => { }
```
`AnyFunc` would be (more or less) the same as `noop` proposed in this PR.

2. Do we need: `..._args: unknown[]` at all? isn't empty arg list would work the same?

CC: @quezak

quezak:
1. `() => void` and () => any` is a different noop. I'd say both are useful in different scenarios, so now I don't know which one should be named just `noop` :sweat_smile: Though I liked the `void` one better, what are the cases where you can't use it and can use the `any` noop?

// And if we're adding "different noops", there's also the "identity noop mapping"  `const id = <T>(x: T) => T` :D

Your proposals are OK, though I'd export them as `ANY_VALUE` (it's a constant, not a type/class/etc) and "some kind of noop" (definitely not capitalized, it's not a type).

2. Yeah, explicit args shouldn't be required. @leaumar did you have any usecase that doesn't work if the function is just `() => {}`?

leaumar:
Some thoughts of mine:
> returning `any` instead of `void`

`noop-ts` does this, and IMO the `any` keyword should be avoided as much as possible because it can sneakily introduce bugs by removing type checking.

> Do we need `..._args`

I _thought_ we did, but I just tried and it seems typescript allows you to assign a parameterless function to a variable typed as a function _with_ parameters, so for that part I guess we don't.

You'd still need it if you want to invoke the function explicitly with some parameters though, e.g. to swallow an error. Now I know it was mentioned before that that comes down to circumventing linters, but I don't entirely agree with that. It is in some coding styles and project configs, but it isn't in others. I for one find `catch (expected) { noop(expected); }` for example a nicely explicit, short, refactor-proof pattern that isn't dependent on configuring any code inspections tools to let it pass.
I don't see any downsides to letting it accept arguments anyway, so we may as well in order to cover more bases, no?

> And if we're adding "different noops"

Perhaps instead of following the simple old `noop` trope, `ts-essentials` could bring something more versatile, like this? (Names just quick illustrative spitballing)
```typescript
// idle, inert, dummy, placeholder, something like that
export const noble = {
  empty = () => {},
  consume = (..._args) => {},
  identity = function<T>(x: T) {return xl},
  // more variations
};
```

quezak:
> You'd still need it if you want to invoke the function explicitly with some parameters 

yeah, for me that's an arugment to keep the unused args there

> `catch (expected) { noop(expected); } `

For me this is not explicit enough ;) but it would be nice with `import { noop as ignoreResult }` or something.

I think let's for now add a plain `noop` as propsed originally, returning void with ignored args, and think later what more is needed.

leaumar:
I guess for the future, if functions were to be added and grouped into an overarching object, setting up a double export like `export const noop = dummy.fn` for backward compatibility wouldn't be too bad.

krzkaczor:
@leaumar @quezak I love this discussion, thanks!

> I think let's for now add a plain noop as propsed originally, returning void with ignored args, and think later what more is needed.

AGREED. Let's just merge it as a "sane" noop. We will continue with what I proposed on the appropriate issue.

krzkaczor:
@all-contributors please add @leaumar for code, ideas, doc

krzkaczor:
Thanks again @leaumar !

Published as `ts-essentials@6.0.3`

leaumar:
Thanks for merging :) Will be using it right away

PR #140, opened 2020-03-30
Add Head and Tail types

quezak:


krzkaczor:
Thanks! Released as 📦 `ts-essentials@6.0.4`

PR #142, opened 2020-03-31
Proposal: format dist/*.d.ts files

quezak:
Rationale: improved DX -- if `dist/types.d.ts` is formatted, using a ts-essentials type you can click "go to declaration" in the editor and examine the details, without leaving your editor to check the docs etc. The currently shipped compiled version doesn't include whitespace so it's unreadable.

Drawbacks: the additional whitespace bumps the `types.d.ts` file size by ~10%.
```
   0[quezak@...]~/src/ts-essentials>: ls -lh dist/*.d.ts
-rw-r--r-- 1 quezak quezak  229 03-31 11:26 dist/functions.d.ts
-rw-r--r-- 1 quezak quezak   54 03-31 11:26 dist/index.d.ts
-rw-r--r-- 1 quezak quezak 9.9K 03-31 11:26 dist/types.d.ts
   0[quezak@...]~/src/ts-essentials>: yarn formatDeclarations
...
   0[quezak@...]~/src/ts-essentials>: ls -lh dist/*.d.ts
-rw-r--r-- 1 quezak quezak 227 03-31 11:27 dist/functions.d.ts
-rw-r--r-- 1 quezak quezak  54 03-31 11:26 dist/index.d.ts
-rw-r--r-- 1 quezak quezak 11K 03-31 11:27 dist/types.d.ts
```

Question: is the readability bonus worth the increased file size?

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,8 @@
   "license": "MIT",
   "scripts": {
     "build": "rm -rf ./dist && tsc -p tsconfig.prod.json --outDir ./dist",
-    "prepublishOnly": "yarn test && yarn build",
+    "formatDeclarations": "prettier --ignore-path *.js --write dist/*.d.ts",
+    "prepublishOnly": "yarn test && yarn build && yarn formatDeclarations",
     "test": "prettier -c **/*.ts && tsc --noEmit",
     "test:fix": "prettier --write **/*.ts && tsc --noEmit"
   },


krzkaczor:
For me it makes sense 👍 

Sidenote: I wonder shouldn't we just copy types.ds and rename It to d.ts - what's compilation is doing here anyway?

quezak:
> Sidenote: I wonder shouldn't we just copy types.ds and rename It to d.ts - what's compilation is doing here anyway?

I guess tsc just copies the type declarations from `.ts` files to `.d.ts` files -- but I'd keep it as is, to avoid complicating the build process with custom hacks.

krzkaczor:
Thanks! Released as 📦 `ts-essentials@6.0.4`

PR #149, opened 2020-05-25
Add ElementOf

### LINKED ISSUE #148: add `ElementOf<T extends any[]>`
We already have `ValueOf<T>` for objects, I thought it would be nice to have a similar helper for array element type, since `ValueOf` with arrays includes more than just the elements [1].

Why I'm doing this via an issue and not a PR is because I have two seemingly equal implementations, does anyone know any advantages of either of those? (like performance, better results for edge cases, better error messages, etc)

```typescript
type ElementOf<T extends any[]> = T extends (infer ET)[] ? ET : never;
type ElementOf<T extends any[]> = T[number];
```

[1] 
```typescript
const1 a1 = [true, false];
type A1ET = ValueOf<typeof arr1>`;

// results:
type A1ET = number | boolean | (() => string) | (() => string) | (() => boolean | undefined) | ((...items: boolean[]) => number) | {
    (...items: ConcatArray<boolean>[]): boolean[];
    (...items: (boolean | ConcatArray<boolean>)[]): boolean[];
} | ... 23 more ... | ((searchElement: boolean, fromIndex?: number | undefined) => boolean)
```

quezak:
Closes #148

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,6 +50,7 @@ npm install --save-dev ts-essentials
   - [Tuple constraint](#Tuple-constraint)
   - [Exhaustive switch cases](#Exhaustive-switch-cases)
   - [ValueOf type](#ValueOf-type)
+  - [ElementOf type](#ElementOf-type)
   - [AsyncOrSync type](#AsyncOrSync-type)
   - [Assertions](#Assertions)
   - [Exact](#Exact)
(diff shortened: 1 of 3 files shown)


quezak:
^ (force-pushed because of accidental readme reformat done by a markdown plugin)

PR #151, opened 2020-05-27
Do not require using all SafeDictionary keys

akwodkiewicz:
Presenting:
```typescript
type OptionalDictionary<T, K extends string | number = string> = { [key in K]?: T };
```
a weaker version of `SafeDictionary<T, K>` -- if the `K` type is **not infinite** you **do not** have to use all its union elements.

```typescript
enum MetadataKeys = {
  NAME = 'NAME',
  AUTHOR = 'AUTHOR',
  TIMESTAMP = 'TIMESTAMP',
}
const metadata: OptionalDictionary<string, MetadataKeys> = {
  AUTHOR: 'akwodkiewicz@gmail.com',
}

// whereas:

const badMetadata: SafeDictionary<string, MetadataKeys> = {
  AUTHOR: 'akwodkiewicz@gmail.com',
}
/// Error:
/// Type '{ AUTHOR: string; }' is missing the following properties
/// from type 'Dictionary<string | undefined, MetadataKeys>': NAME, TIMESTAMP
```
There is no optional version of _non-safe_ `Dictionary`, because the optionality of keys makes the return type implicitly `T | undefined`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,6 +96,13 @@ type stringDictValues = DictionaryValues<typeof stringDict>;
 // When building a map using JS objects consider using SafeDictionary
 const safeDict: SafeDictionary<number> = {}
 const value: number | undefined = safeDict['foo']
+
+// ...or OptionalDictionary
+type ConfigKeys = 'LOGLEVEL' | 'PORT' | 'DEBUG';
+const optionalDict: OptionalDictionary<string, ConfigKeys> = {
+  LOGLEVEL: 'INFO',
+}
+const port: string | undefined = optionalDict['PORT'];
 ```
 
 ### Deep* wrapper types
--- a/README.md
+++ b/README.md
@@ -101,12 +101,20 @@ type stringDictValues = DictionaryValues<typeof stringDict>;
 const safeDict: SafeDictionary<number> = {}
 const value: number | undefined = safeDict['foo']
 
-// ...or OptionalDictionary
+// With SafeDictionary you don't need to use all of the sub-types of a finite type.
+// If you care about the key exhaustiveness, use a regular Dictionary.
 type ConfigKeys = 'LOGLEVEL' | 'PORT' | 'DEBUG';
-const optionalDict: OptionalDictionary<string, ConfigKeys> = {
-  LOGLEVEL: 'INFO',
+const configSafeDict: SafeDictionary<number, ConfigKeys> = {
+  LOGLEVEL: 2,
 }
-const port: string | undefined = optionalDict['PORT'];
+const maybePort: number | undefined = configSafeDict['PORT'];
+
+const configDict: Dictionary<number, ConfigKeys> = {
+  LOGLEVEL: 2,
+  PORT: 8080,
+  DEBUG: 1,
+}
+const port: number = configDict['PORT'];
 ```
 
 ### Deep\* wrapper types
--- a/lib/types.ts
+++ b/lib/types.ts
@@ -24,6 +24,8 @@ export type DictionaryValues<T> = T extends Dictionary<infer U> ? U : never;
  * because using `Dictionary<T, 'a' | 'b'>` already enforces that all of the keys are present.
  */
 export type SafeDictionary<T, K extends string | number = string> = Dictionary<T | undefined, K>;
+/** Like SafeDictionary, but not require key exhaustiveness */
+export type OptionalDictionary<T, K extends string | number = string> = { [key in K]?: T };
 
 /** Like Partial but recursive */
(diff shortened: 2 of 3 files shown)


krzkaczor:
I see what you mean. What if, instead of adding new type, we tweak `SafeDictionary`? 

Something like:
```typescript
type SafeDictionary<T, K extends string | number = string> = { [key in K]?: T | undefined };

It shouldn't be a breaking change 🤔

**Comment by akwodkiewicz** [comment]
Oh, it **would** be a breaking change -- we rely on `SafeDictionary`'s exhaustiveness to make sure that expanding existing enums is handled properly everywhere. If you change the implementation, then we would lose all the type safety.

**Comment by krzkaczor** [comment]
> Oh, it would be a breaking change -- we rely on SafeDictionary's exhaustiveness to make sure that expanding existing enums is handled properly everywhere.

Can u create a small example to demonstrate how is this a breakin change, I am not sure if I understand? The way I think about it, with the current `SafeDictionary` you have to provide explicit `undefined`, if we tweak it then it's implicitly `undefined` (it's not existing property). It only makes a difference if you use something like `in` operator. Do you?

**Comment by akwodkiewicz** [comment]
Sure! Let's say we've got:

```typescript
enum Integrations {
  ACEBOOK = 'ACEBOOK',
  BOOGLE = 'BOOGLE',
  CRITTER = 'CRITTER',
};
```

and in some other file, there are 2 implementations of our integration IDs dictionaries:

```typescript
export const myIntegrationIdDict: SafeDictionary<string, Integrations> = {
  ACEBOOK: 'id-1234',
  BOOGLE: 'abc-def',
  CRITTER: '012345'
};

export const myUnsafeIntegrationIdDict: OptionalDictionary<string, Integrations> = {
  ACEBOOK: 'id-1234',
  BOOGLE: 'abc-def',
  CRITTER: '012345'
};
```

One lovely Wednesday afternoon you decide to add the 4th integration:

```diff
enum Integrations {
  ACEBOOK = 'ACEBOOK',
  BOOGLE = 'BOOGLE',
  CRITTER = 'CRITTER',
+ DING = 'DING' // let's see how this will work!
};
```

Now it's obvious you should've added the DING integration ID to the dictionaries, but at the time you simply forgot. Depending on which implementation you used, the compiler would've warned you about the mistake:

```typescript
export const myIntegrationIdDict: SafeDictionary<string, Integrations> = {
  ACEBOOK: 'id-1234',
  BOOGLE: 'abc-def',
  CRITTER: '012345'
};
/// ERROR:
/// Property 'DING' is missing in type
/// '{ ACEBOOK: string; BOOGLE: string; CRITTER: string; }'
/// but required in type 'Dictionary<string | undefined, Integrations>'. ts(2741)



export const myUnsafeIntegrationIdDict: OptionalDictionary<string, Integrations> = {
  ACEBOOK: 'id-1234',
  BOOGLE: 'abc-def',
  CRITTER: '012345'
};
// However this one, has no errors at all!
```

If you use the `OptionalDictionary` implementation in place of `SafeDictionary`, then you lose type safety which would forbid you from using incomplete dictionaries.

krzkaczor:
Okay, I get you but why would you use `SafeDictionary` instead of `Dictionary` in this example? 

I still feel like we shouldn't add yet another type but rather do a breaking change and explain in readme when should you use one or the other (including mentioning enums)

quezak:
`SafeDictionary` was introduced recently enough that I think we can make it a breaking change.
Maybe it's not obvious from the readme, and only briefly mentioned in the comment, `SafeDictionary` was intended to be used only with infinite types. It was even introduced with only the value parameter, and the key was always `string` -- the second param was actually added by me because I needed `SafeDictionary<T, number>` in one place :D so, all in all, it's not _that big_ of a breaking change.

akwodkiewicz:
> `SafeDictionary` was intended to be used only with infinite types

If that's the case then alright, we simply need to emphasize this in README.

akwodkiewicz:
I changed the `SafeDictionary` implementation to the one from my proposal for `OptionalDictionary`.

akwodkiewicz:
I'd like to be included in the contributor list :slightly_smiling_face:

quezak:
ah yes, I thought you're already there :sweat_smile: 

@all-contributors please add @akwodkiewicz  for code, doc, ideas

krzkaczor:
Great work folks! Note: I merged it to a separate branch and I will treat it as a breaking change. 

I want to include breaking changes regarding `Opaque` types in version 7 too. https://github.com/krzkaczor/ts-essentials/issues/158

krzkaczor:
Released as `ts-essentials@7.0.0`

PR #153, opened 2020-05-31
fix #118, keep optional flag in DeepOmit members

### LINKED ISSUE #118: DeepOmit changes optional fields to nullable?
- Example:
```

type Test = {
  id: string;
  age: number;
  name?: string
}
type OmittedTest = DeepOmit<Test, {age: never}>
```
Screen Shot 2020-01-06 at 17 05 32

- My question is: why does the deepOmit turn optional fied `name?:string` to `name:string|undefined`

bbarry:
While I feel this is a typescript bug in that the compiler loses attributes on the members of a type when member names are filtered, this probably should still get worked around in ts-essentials.

The resulting DeepOmit type is not exactly the same because the members are in a different order, but I am not sure how to possibly test member ordering within Typescript.

I also made RequiredKeys and OptionalKeys exported here because I have found good utility out of them in other projects.

### DIFF
--- a/README.md
+++ b/README.md
@@ -506,7 +506,7 @@ Gets keys of an object which are optional.
 type T = {
   a: number;
   b?: string;
-  c: string;
+  c: string | undefined;
   d?: string;
 };
 type Result = OptionalKeys<T>;
@@ -522,7 +522,7 @@ Gets keys of an object which are required.
 type T = {
   a: number;
   b?: string;
-  c: string;
+  c: string | undefined;
   d?: string;
 };
 type Result = OptionalKeys<T>;
--- a/README.md
+++ b/README.md
@@ -17,7 +17,8 @@
 npm install --save-dev ts-essentials
 ```
 
-👉 We require `typescript>=3.7`. If you're looking for support for older TS versions use `ts-essentials@3` (for 3.6>=) or `ts-essentials@2` instead.
+👉 We require `typescript>=3.7`. If you're looking for support for older TS versions use `ts-essentials@3` (for 3.6>=)
+or `ts-essentials@2` instead.
 
 ## What's inside?
 
@@ -25,7 +26,7 @@ npm install --save-dev ts-essentials
 - [What's inside?](#Whats-inside)
   - [Basic](#Basic)
   - [Dictionaries](#Dictionaries)
-  - [Deep* wrapper types](#Deep-wrapper-types)
+  - [Deep\* wrapper types](#Deep-wrapper-types)
     - DeepPartial
     - DeepRequired
     - DeepReadonly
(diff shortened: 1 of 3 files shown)


lib/types.ts
```diff
@@ -205,10 +211,21 @@ export type DeepOmit<T extends DeepOmitModify<Filter>, Filter> = T extends Built
   ? ItemType extends DeepOmitModify<Filter>
     ? Promise<DeepOmit<ItemType, Filter>>
     : T
-  : { [K in Exclude<keyof T, keyof Filter>]: T[K] } &
+  : { [K in Exclude<OptionalKeys<T>, keyof Filter>]+?: T[K] } &
```
quezak, line 214:
```suggestion
  // explicitly mentioning optional properties, to work around TS making them required
  // see https://github.com/krzkaczor/ts-essentials/issues/118
  : { [K in Exclude<OptionalKeys<T>, keyof Filter>]+?: T[K] } &
```

krzkaczor:
Hmm I am not sure why linter fails. Do you use yarn? (maybe you got a wrong version of prettier installed). Anyway, I am gonna merge this manually.

bbarry:
It doesn't like the comment on multiple lines. I just opened and ran prettier again and it changed the comment to one line.

bbarry:
Looks like https://github.com/prettier/prettier/issues/6352.

The comment moves around a few times when you save. Maybe if we want to keep the comment, extract a helper type for the ternary operation here?

README.md
```diff
@@ -487,11 +493,43 @@ type T = {
...
 
+### OptionalKeys
+
+Gets keys of an object which are optional.
+
+```typescript
+type T = {
+  a: number;
+  b?: string;
+  c: string;
```
quezak, line 509: we could make one key "undefinable", so it's clearer for the unsure readers that this type works on the keys not on value type
```suggestion
  c: string | undefined;
```

README.md
```diff
@@ -487,11 +493,43 @@ type T = {
...
+
+### RequiredKeys
+
+Gets keys of an object which are required.
+
+```typescript
+type T = {
+  a: number;
+  b?: string;
+  c: string;
```
quezak, line 525:
```suggestion
  c: string | undefined;
```

lib/types.ts
```diff
@@ -204,11 +212,23 @@ export type DeepOmit<T extends DeepOmitModify<Filter>, Filter> = T extends Built
   : T extends Promise<infer ItemType>
   ? ItemType extends DeepOmitModify<Filter>
     ? Promise<DeepOmit<ItemType, Filter>>
-    : T
-  : { [K in Exclude<keyof T, keyof Filter>]: T[K] } &
+    : T // explicitly mentioning optional properties, to work around TS making them required
+  : // see https://github.com/krzkaczor/ts-essentials/issues/118
```
quezak, line 216: we can move this comment to the top of `DeepOmit` if prettier has issues with it

quezak:
@krzkaczor I think we're good to go now

quezak:
@all-contributors please add @bbarry for code, doc

krzkaczor:
Thanks guys! 

I will ensure that prettier always formats readme and then I will make a release.

krzkaczor:
Released as `ts-essentials@6.0.6`

PR #186, opened 2020-12-09
DeepReadonly: Type 'unknown' is not assignable to type 'Readonly<unknown>'

Julusian:
```
type Test1 = {
	abc: unknown
}

const val: Test1 = { abc: '' }
const val2: DeepReadonly<Test1> = val
```
Gives the error:
```
error TS2322: Type 'Test1' is not assignable to type '{ readonly abc: Readonly<unknown>; }'.
  Types of property 'abc' are incompatible.
    Type 'unknown' is not assignable to type 'Readonly<unknown>'.
```

This lets it translate the type cleanly across

### DIFF
--- a/lib/types.ts
+++ b/lib/types.ts
@@ -1,5 +1,5 @@
 /** Essentials */
-export type Primitive = string | number | boolean | bigint | symbol | undefined | null | unknown;
+export type Primitive = string | number | boolean | bigint | symbol | undefined | null;
 export type Builtin = Primitive | Function | Date | Error | RegExp;
 export type IsTuple<T> = T extends [infer A]
   ? T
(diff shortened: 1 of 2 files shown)


quezak:
I don't think this is a valid solution -- `unknown` is not a primitive type, same as `any` isn't. And while my concern initially was only about the naming / meaning of the "primitive type" phrase, it looks like the tests stopped working :)

Julusian:
yeah once I saw the tests failing I started to have a deeper look locally. Ill unmark the pr as a draft once I have a solution which passes in the tests

Julusian:
I don't have any ideas right now on how to match unknown in a complex interface.
For now I have updated this to include an property of type unknown in the test cases, to show which types have problems, and I shall return if I can find anything on how to match unknown

PR #187, opened 2021-01-22
Make build script cross-platform

leaumar:
`rm -rf` is not cross-platform, please merge this to enable windows devs to work on this repo. The `&&`, though not windows-shell-compatible, is fine because yarn specifically makes it work.


krzkaczor:
Thanks - that's useful.

PR #188, opened 2021-01-22
Add editorconfig

leaumar:
Adding a `.editorconfig` to your repo makes all the popular editors automatically respect some formatting rules regardless of personal settings to reduce first-time-setup friction for contributors. Indentation of 2 for example is not default with many people, making edits more tedious. Prettier does not help in this regard.

~The `end_of_line` may look contestable but no windows tool other than literally notepad.exe has issues with `lf`, and it removes any reliance on git's wonky auto-crlf (which in turn depends on careful git configuration) — uniformation beats configuration. This may work more nicely with an accompanying `.gitattributes` to force-disable git's autocrlf to prevent constant false change tracking on changed line-endings, but let's take 1 step at a time.~

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,9 @@
+[*]
+charset = utf-8
+indent_size = 2
+indent_style = space
+insert_final_newline = true
+trim_trailing_whitespace = true
+
+[*.md]
+trim_trailing_whitespace = false


krzkaczor:
Do you know how this works with prettier? Indentation is handled by prettier i don't know about the rest tbh.

leaumar:
Combining the two has always worked perfectly for me. Prettier only works when invoked manually or if you actively set up your IDE with it when you import the project. Even then, I've never known the "apply style settings based on prettier config" in my IDE to actually work. The editorconfig applies immediately and mainly fixes your indentation without any intervention. Not having this file makes my IDE fight with the existing code indents.

I will trim the config down to just fixing indents though, I figure now that the EOL doesn't need to be forced for any reason (and might end up fighting people's autocrlf).

krzkaczor:
Okay so it seems like editor config can be useful for all files while prettier only works for some of them. I agree that it makes sense to have both. Thanks!

PR #189, opened 2021-01-22
Fix invalid sourcemap reference in published files

### LINKED ISSUE #169: Bundling bug: missing sourcemap
Every time I build some projects of mine, I get the following warning on build:
> $ parcel build src/*.* --out-dir extension --no-autoinstall
> ‼  Could not load source file "../lib/index.ts" in source map of "../node_modules/ts-essentials/dist/index.js".
> ‼  Could not load source file "../lib/functions.ts" in source map of "../node_modules/ts-essentials/dist/functions.js".

`index.js.map` specifies `sources":["../lib/index.ts"]`, but the `lib` folder is indeed not included in node_modules:
image

I don't know how sourcemaps and npm artifact publishing work, but I guess either `lib` needs to be published with the rest, or the sourcemap needs to be inlined into this file.

Idem for `functions.js` of course. It shouldn't make a difference but I use yarn 1.22.4.

Edit: come to think of it, there probably isn't much to put into a js sourcemap as it's 99% compiled-away types, but then at least there should be a nigh-empty sourcemap instead of a missing file so build tools don't throw errors when resolving it.

leaumar:
Fixes #169 

I tried to set `inlineSources` first to simply change the invalid (unpublished) source file reference to an inline source and keep the whole sourcemap as a separate file, but ~despite what [the docs](https://www.typescriptlang.org/tsconfig#inlineSources) claim, this option didn't change the output,~ the same reference to `../lib` was still emitted.

Lacking a fix for that, the next step was to just inline it all into the js file.

PS: this PR is the reason for #187 and #188

### DIFF
--- a/tsconfig.prod.json
+++ b/tsconfig.prod.json
@@ -1,5 +1,10 @@
 {
   "extends": "./tsconfig.json",
+  "compilerOptions": {
+    "sourceMap": false,
+    "inlineSourceMap": true,
+    "inlineSources": true
+  },
   "include": ["lib/**/*.ts"],
   "exclude": ["node_modules", "example"]
 }


krzkaczor:
Can you move these changes to `tsconfig.prod.json`? so it's used only for production builds?

leaumar:
Sure

krzkaczor:
Sorry to keep you waiting for so long. 

I just realized that perhaps we should not ship sourcemaps at all this PR makes bundle size explode?
> Build size (runtime) — Change: +19.17KB (+2280.02%) Total: 20.01KB

I understand that in the past they were misconfigured maybe now they should just be explicitly turned off? I don't see a reason to ship them, and this shall fix bundler problems too. WDYT?

leaumar:
I don't think bundle size is a real issue. Most real downstream projects churn everything through their own bundler to get the leanest form of all their code anyway.

But if trimming it anyway for some minor use cases comes at no functional cost, then sure. So the question is what do we lose if we omit sourcemaps, especially in a typescript one that's 99% compiler-omitted types? I haven't used ts-essentials much yet and haven't needed to debug around it yet, but I have often dived into library code while debugging nearby issues and been thankful for the sourcemaps to help make sense of the impl behind the API.

So I'd say it comes down to what value they are in debugging (which I do consider an important usecase), but I can't answer that from experience.

krzkaczor:
My reasoning is that the "runtime" part is extremely small - just a couple of simple functions and that's why it doesn't need sourcemaps. Also I am not really sure if you want source maps lying around (what about users who don't bundle their code?).

leaumar:
Well they help in debugging, even for typescript as you can load ts sources into e.g. a browser debugger to go read the definition of a type. I imagine omitting sourcemaps would make debugging ts sources harder since you couldn't click-through on stuff anymore. But I don't have enough experience debugging ts libs to back this up with.

Do people who don't bundle their code care about 1 file of a few kilobytes among what's already dozens or hundreds of megabytes worth of deployment luggage?

krzkaczor:
I think it's fine to merge. Sorry that it took so long and thanks for your contribution :)

leaumar:
Thanks 👍

PR #210, opened 2021-08-03
docs: 🔄 update README for Opaque type

Beraliv:
You don't need to cast it to `unknown` or `any` from base type to opaque

Example is provided here – https://tsplay.dev/Wy5XgW

Forgot when implemented the PR – https://github.com/krzkaczor/ts-essentials/pull/200

CC @krzkaczor

### DIFF
--- a/README.md
+++ b/README.md
@@ -627,7 +627,7 @@ function makePositiveNumber(n: number): PositiveNumber {
   if (n <= 0) {
     throw new Error(`Value ${n} is not positive !`);
   }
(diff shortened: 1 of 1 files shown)


krzkaczor:
I don't fully understand why `any` cast is not required now but it was needed in the past 🤔 is it because of your PR? or some tsc changes

Beraliv:
> I don't fully understand why `any` cast is not required now but it was needed in the past 🤔 is it because of your PR? or some tsc changes

Yeah, it's because for this implementation you extend base type and cast is allowed

krzkaczor:
Thanks!

PR #213, opened 2021-08-13
feat: ➕ add MergeN

### LINKED ISSUE #209: Can merge be used with an array of N types to be merged?
If I have an array like:

```ts
const rules = [
    {} as { foo: 1 | 2 | 3},
    {} as { bar: 4 | 5 | 6},
]
```

I basically want to do:

```ts
type Rules = Merge<typeof rules[0], typeof rules[1]>
```

In order to get:

```ts
type Rules = {
  foo?: 1 | 2 | 3
  bar?: 4 | 5 | 6
}
```

But for N rules in the array.

Is this possible?

Beraliv:
Fixes #209

### DIFF
--- a/lib/types.ts
+++ b/lib/types.ts
@@ -284,6 +284,13 @@ export type NonEmptyObject<T extends {}> = keyof T extends never ? never : T;
 /** Merge 2 types, properties types from the latter override the ones defined on the former type */
 export type Merge<M, N> = Omit<M, keyof N> & N;
 
+type _MergeN<T extends readonly any[], Result> = T extends readonly [infer Head, ...(infer Tail)]
+  ? _MergeN<Tail, Merge<Result, Head>>
+  : Result;
+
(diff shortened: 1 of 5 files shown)


lib/types.ts
```diff
@@ -284,6 +284,13 @@ export type NonEmptyObject<T extends {}> = keyof T extends never ? never : T;
 /** Merge 2 types, properties types from the latter override the ones defined on the former type */
 export type Merge<M, N> = Omit<M, keyof N> & N;
 
+type _MergeN<T extends readonly any[], Result> = T extends readonly [infer Head, ...(infer Tail)]
```
Beraliv, line 287: Ah, that's only available since TypeScript@4.0 – https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-0.html

krzkaczor (approved):

PR #214, opened 2021-08-15
feat: 🚀 improve IsTuple

Beraliv:
Can check for tuples of all size now, not until 5 elements


lib/types.ts
```diff
@@ -1,17 +1,7 @@
...
-  : T extends [infer A, infer B]
-  ? T
-  : T extends [infer A, infer B, infer C]
-  ? T
-  : T extends [infer A, infer B, infer C, infer D]
-  ? T
-  : T extends [infer A, infer B, infer C, infer D, infer E]
-  ? T
-  : never;
+export type IsTuple<T> = T extends any[] ? (any[] extends T ? never : T) : never;
```
Beraliv, line 4: I found that check based on my article – https://blog.beraliv.dev/2021-03-26-typed-get/#33-tuples

I built the table for that and distinguished the cases where array and where tuple

krzkaczor (approved):
Nice work!

test/index.ts
```diff
@@ -233,7 +232,8 @@ function testDeepPartial() {
     Assert<IsExact<DeepPartial<Set<string>>, Set<string>>>,
     Assert<IsExact<DeepPartial<Set<number[]>>, Set<number[]>>>,
     Assert<IsExact<DeepPartial<ReadonlySet<string>>, ReadonlySet<string>>>,
-    Assert<IsExact<DeepPartial<[]>, never[]>>,
+    Assert<IsExact<DeepPartial<[]>, []>>,
```
Beraliv, line 235: So now, as `[]` is inferred differently, we will get `[]` instead of `never[]` which is in general fine

PR #175, opened 2020-08-09
add DeepPick type

cyberbiont:
this adds DeepPick type and tests for it, see
https://github.com/krzkaczor/ts-essentials/issues/168

DeepOmitModify helper class was  changed to FilterModify since now it is used by both DeepOmit and DeepPick types.
I also changed `Filter` type argument to `F` for clarity.

PR #224, opened 2021-08-19
feat: ➕ add DeepPick type

Beraliv:
It's the extension of #175 and will be merged to `master` once it's fixed for all the tests

TODO:
- [x] objects
- [x] maps
- [x] sets
- [x] arrays
- [x] tuples
- [x] promise 
- [x] update prettier to support TypeScript@4.1 (done in #269)

### DIFF
--- a/lib/types.ts
+++ b/lib/types.ts
@@ -203,114 +203,114 @@ export type PickKeys<T, P> = Exclude<keyof PickProperties<T, P>, undefined>;
 /** Recursively omit deep properties */
 // explicitly mentioning optional properties, to work around TS making them required
 // see https://github.com/krzkaczor/ts-essentials/issues/118
-export type DeepOmit<T extends DeepModify<F>, F> = T extends Builtin
+export type DeepOmit<T extends DeepModify<Filter>, Filter> = T extends Builtin
   ? T
   : T extends Map<infer KeyType, infer ValueType>
-  ? ValueType extends DeepModify<F>
-    ? Map<KeyType, DeepOmit<ValueType, F>>
+  ? ValueType extends DeepModify<Filter>
+    ? Map<KeyType, DeepOmit<ValueType, Filter>>
     : T
   : T extends ReadonlyMap<infer KeyType, infer ValueType>
-  ? ValueType extends DeepModify<F>
-    ? ReadonlyMap<KeyType, DeepOmit<ValueType, F>>
+  ? ValueType extends DeepModify<Filter>
+    ? ReadonlyMap<KeyType, DeepOmit<ValueType, Filter>>
     : T
   : T extends WeakMap<infer KeyType, infer ValueType>
-  ? ValueType extends DeepModify<F>
-    ? WeakMap<KeyType, DeepOmit<ValueType, F>>
+  ? ValueType extends DeepModify<Filter>
+    ? WeakMap<KeyType, DeepOmit<ValueType, Filter>>
     : T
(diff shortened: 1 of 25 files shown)


lib/types.ts
```diff
@@ -253,23 +253,80 @@ export type DeepOmit<T extends DeepOmitModify<Filter>, Filter> = T extends Built
...
+          [K in Extract<RequiredKeys<T>, keyof Filter>]: Filter[K] extends true
+            ? T[K]
+            : T[K] extends DeepModify<Filter[K]>
+            ? DeepPick<T[K], Filter[K]>
+            : never;
+        },
+        never
+      >;
+
+type DeepModify<T> =
```
Beraliv, line 319: `DeepModify` should be named `Skeleton` as we want to have all possible combinations of keys without thinking about values

Can experiment with it when cover everything with tests and fix it

cyberbiont: Those tests that are failing in your TS Playground deal with _optional_ properties of ComplexNestedObject.
And yes, it doesn't work this way at the moment (I don't remember exactly why, but there was some reason, when I have time maybe I'll look into it).

Anyway, the DeepOmit type, that is already present in ts-essentials from long ago, works the same way, i.e. you cannot deep omit optional properties (you can try and test it). 
DeepPick feature was designed as complementary type for DeepOmit and as such it works just fine.
This means that you can release this feature, and we'll see later if something can be done about this particular case.

Beraliv: > All those tests that are failing in your TS Playground deal with optional properties of ComplexNestedObject.
And yes, it doesn't work this way at the moment

Yeah, I know, I will try to fix it here, that's why I created another MR

> Anyway, the DeepOmit type, that is already present in ts-essentials from long ago, works the same way, i.e. you cannot deep omit optional properties (you can try and test it).

That's true, but as we're preparing for the release, I would improve it (as I also have time on the weekend) because anyway we touch it and change

> Then I'll make a new PR that fixes this behaviour in both DeepPick and DeepOmit.

You can work on top of this PR if it's fine for you
I didn't mean to change anything else except for that

lib/types.ts
```diff
@@ -271,23 +273,74 @@ export type DeepOmit<T extends DeepOmitModify<Filter>, Filter> = T extends Built
...
+  ? ItemType extends DeepModify<Filter>
+    ? Array<DeepPick<ItemType, Filter>>
+    : T
+  : T extends Promise<infer ItemType>
+  ? ItemType extends DeepModify<Filter>
+    ? Promise<DeepPick<ItemType, Filter>>
+    : T
+  : Filter extends Record<string, unknown>
+  ? {
+      [K in keyof Filter & OptionalKeys<T>]?: Filter[K] extends true ? T[K] : DeepPick<T[K], Filter[K]>;
```
Beraliv, line 320: TODO:

1. Add test cases with `never`
2. Add `IsNever` check after merge of https://github.com/krzkaczor/ts-essentials/pull/251

Beraliv: - [x] Added test cases for `never` filter
- [x] Don't need to unify check for the filter as it's already mentioned once in `DeepModify`

lib/types.ts
Beraliv: Done here, will keep `DeepModify` for now for only `DeepPick`

`DeepOmit` still uses `DeepOmitModify` to not escalate errors

krzkaczor (approved):

PR #225, opened 2021-08-20
Deep pick

cyberbiont:
partially fixes optional properties problem in Deep Pick / Omit.


cyberbiont:
Here is my attempt to fix it.
It works only with 'never' type for filter again, so I modified all tests to use `never`.

As a side-effect, `optionalProp?: undefined` will be still preset in resulting omitted type. (not a big problem I hope because usually nobody explicitly assigns properties with value 'undefined').

Also, currently it doesn't work for optional properties, nested inside optional properties... This probably can be fixed but I don't have time to dive in it now.

Beraliv:
> Here is my attempt to fix it.
> It works only with 'never' type for filter again, so I modified all tests to use `never`.
> 
> As a side-effect, `optionalProp?: undefined` will be still preset in resulting omitted type. (not a big problem I hope, because usually nobody explicitly assigns properties with value 'undefined').
> 
> Also, currently it doesn't work for optional properties, nested inside optional properties... This probably can be fixed but I don't have time to dive in it now.
> 
> Maybe you can come up with better solution.

Thank you for creating PR, I will pay some attention on the weekend, maybe come up with something that can help fixing it

README.md
```diff
@@ -27,55 +27,48 @@ If you use any [functions](https://github.com/krzkaczor/ts-essentials/blob/maste
...
-  - [NonNever](#NonNever)
-  - [Merge](#Merge)
-  - [MergeN](#MergeN)
-  - [MarkRequired](#MarkRequired)
-  - [MarkOptional](#MarkOptional)
-  - [ReadonlyKeys](#ReadonlyKeys)
-  - [WritableKeys](#WritableKeys)
-  - [OptionalKeys](#OptionalKeys)
-  - [RequiredKeys](#RequiredKeys)
+- [Install](#install)
```
Beraliv, line 30: Can you please revert these changes or push them separately?

That's out of the scope of the changes

lib/types.ts
```diff
@@ -251,7 +251,7 @@ export type DeepOmit<T extends DeepModify<Filter>, Filter> = T extends Builtin
       { [K in Exclude<RequiredKeys<T>, keyof Filter>]: T[K] } &
       OmitProperties<
         {
-          [K in Extract<RequiredKeys<T>, keyof Filter>]: Filter[K] extends true
+          [K in Extract<RequiredKeys<T>, keyof Filter>]: Filter[K] extends never
```
Beraliv, line 254: Given `T extends never` we distribute the type, but not checking that it is `never` (see [Distributive Conditional Types](https://www.typescriptlang.org/docs/handbook/advanced-types.html#distributive-conditional-types))

To check that the type is `never` you need to use this types specifically:

```ts
type IsNever<T> = [T] extends [never] ? true : false;
```

README.md
cyberbiont: reverted. it seems like VS Code Markdown formatter (that comes Markdown All In One extension) doesn't like plain text items like - DeepPartial  inside the table of contents and auto-deletes them. Didn't notice it at first. I disabled the extension, but maybe you should consider to modify those (make them also a links perhaps)

Beraliv:
@cyberbiont I would close it in favour to https://github.com/krzkaczor/ts-essentials/pull/224

What do you think?

Can you please test the implementation that I've completed? Do you have any suggestions?

PR #227, opened 2021-08-21
docs: ➕ add strictNullChecks 📄

### LINKED ISSUE #196: RequiredKeys generates never
```
interface Test {
  opt?: string,
  req1: string,
  req2: string,
}

const key: RequiredKeys<Test> = 'req1';
```
Should have no errors.
But generates the error:
Type 'string' is not assignable to type 'never'.ts(2322)

The type `RequiredKeys<Test>` should return `"req1" | "req2"`. But instead generates `never`.

Beraliv:
Fixes #196

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,6 +20,9 @@ npm install --save-dev ts-essentials
 👉 We require `typescript>=4.0`. If you're looking for support for older TS versions, please have a look at the
 [TypeScript dependency table](https://github.com/krzkaczor/ts-essentials/tree/master#TypeScript-dependency-table)
 
+👉 As we really want types to be stricter, we require enabled
+[strictNullChecks](https://www.typescriptlang.org/tsconfig#strictNullChecks) in your project
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -20,6 +20,9 @@ npm install --save-dev ts-essentials
 👉 We require `typescript>=4.0`. If you're looking for support for older TS versions, please have a look at the
 [TypeScript dependency table](https://github.com/krzkaczor/ts-essentials/tree/master#TypeScript-dependency-table)
 
+👉 As we really want types to be stricter, we require enabled
+[strictNullChecks](https://www.typescriptlang.org/tsconfig#strictNullChecks) in your project
+
```
Beraliv, line 25: * I also can leave the link to playground – https://tsplay.dev/mqvpJW (to reproduce the problem)
* Can mention the issue as well – https://github.com/krzkaczor/ts-essentials/issues/196

krzkaczor (approved):
I think this change is good as it is. I don't think that we need to dive more into details explaining why is that -- it's better to keep it short and concise.

PR #228, opened 2021-08-22
test: 🧪 DeepNullable

Beraliv:
Fixed `DeepNullable` for `ReadonlySet` and `ReadonlyMap`

### DIFF
--- a/test/index.ts
+++ b/test/index.ts
@@ -251,6 +251,56 @@ function testDeepPartial() {
   ];
 }
 
+function testDeepNullable() {
+  type cases = [
+    Assert<IsExact<DeepNullable<number>, number | null>>,
+    Assert<IsExact<DeepNullable<string>, string | null>>,
+    Assert<IsExact<DeepNullable<boolean>, boolean | null>>,
+    Assert<IsExact<DeepNullable<bigint>, bigint | null>>,
+    Assert<IsExact<DeepNullable<symbol>, symbol | null>>,
+    Assert<IsExact<DeepNullable<undefined>, undefined | null>>,
+    Assert<IsExact<DeepNullable<null>, null>>,
+    Assert<IsExact<DeepNullable<Function>, Function | null>>,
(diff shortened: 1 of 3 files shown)


test/index.ts
```diff
@@ -251,6 +251,56 @@ function testDeepPartial() {
...
+    Assert<IsExact<DeepNullable<bigint>, bigint | null>>,
+    Assert<IsExact<DeepNullable<symbol>, symbol | null>>,
+    Assert<IsExact<DeepNullable<undefined>, undefined | null>>,
+    Assert<IsExact<DeepNullable<null>, null>>,
+    Assert<IsExact<DeepNullable<Function>, Function | null>>,
+    Assert<IsExact<DeepNullable<Date>, Date | null>>,
+    Assert<IsExact<DeepNullable<Error>, Error | null>>,
+    Assert<IsExact<DeepNullable<RegExp>, RegExp | null>>,
+    Assert<IsExact<DeepNullable<Map<string, boolean>>, Map<string | null, boolean | null>>>,
+    Assert<IsExact<DeepNullable<Map<string, { a: number }>>, Map<string | null, { a: number | null }>>>,
```
krzkaczor, line 268: I am not sure if the behavior to make **keys** of Maps etc, `nullable` is desired. Have you thought about this?

krzkaczor (commented):
Btw. as there is a slight change in the behaviour right now (support readonly maps and sets) this should have a `patch` changeset.

test/index.ts
Beraliv: It's not a change, that's an alignment for `ReadonlyMap`, because it was already working that way for `Map` and `WeekMap`

Same regarding `Set` and `WeekSet` – it was already working this way and I aligned it for `ReadonlySet`

krzkaczor (approved):

krzkaczor (approved):

test/index.ts
Beraliv: Agreed on adding it as an alignment ✅

PR #230, opened 2021-08-22
test: 🧪 more cases for DeepNonNullable

Beraliv:


krzkaczor (commented):
Same comment as for https://github.com/krzkaczor/ts-essentials/pull/228

Are you positive that we want to make keys for map not nullable too?

krzkaczor (approved):

PR #235, opened 2021-08-24
fix: 🧪 DeepRequire

Beraliv:
Make `DeepRequire` as recursive `Require` without removing `null` and `undefined` unexpectedly

For removing `null` and `undefined` `DeepNonNullable` can be used

### DIFF
--- a/test/index.ts
+++ b/test/index.ts
@@ -541,71 +541,94 @@ function testDeepNonNullable() {
 
 function testDeepRequire() {
   type cases = [
-    Assert<IsExact<DeepRequired<number | null | undefined>, number>>,
-    Assert<IsExact<DeepRequired<string | null | undefined>, string>>,
-    Assert<IsExact<DeepRequired<boolean | null | undefined>, boolean>>,
-    Assert<IsExact<DeepRequired<bigint | null | undefined>, bigint>>,
-    Assert<IsExact<DeepRequired<symbol | null | undefined>, symbol>>,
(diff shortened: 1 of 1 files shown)


test/index.ts
```diff
@@ -541,71 +541,94 @@ function testDeepNonNullable() {
...
-    Assert<IsExact<DeepRequired<boolean | null | undefined>, boolean>>,
-    Assert<IsExact<DeepRequired<bigint | null | undefined>, bigint>>,
-    Assert<IsExact<DeepRequired<symbol | null | undefined>, symbol>>,
-    Assert<IsExact<DeepRequired<undefined | null>, never>>,
-    Assert<IsExact<DeepRequired<Function | null | undefined>, Function>>,
-    Assert<IsExact<DeepRequired<Date | null | undefined>, Date>>,
-    Assert<IsExact<DeepRequired<Error | null | undefined>, Error>>,
-    Assert<IsExact<DeepRequired<RegExp | null | undefined>, RegExp>>,
-    Assert<IsExact<DeepRequired<Map<string | null | undefined, boolean | null | undefined>>, Map<string, boolean>>>,
+    Assert<IsExact<DeepRequired<number | null | undefined>, number | null | undefined>>,
```
Beraliv, line 544: The tests are broken from the first commit

You can see that all of the test cases are not valid now – https://github.com/krzkaczor/ts-essentials/runs/3415539355

It will be fixed with the next commit

PR #242, opened 2021-08-28
fix: 🧪 StrictOmit

Beraliv:
1. Allow only objects to be used in `StrictOmit`
2. `StrictOmit` returns `never` for arrays and tuples

### DIFF
--- a/.changeset/long-bears-fetch.md
+++ b/.changeset/long-bears-fetch.md
@@ -1,0 +1,5 @@
+---
+"ts-essentials": patch
+---
+
+Allow objects to be used in `StrictOmit`
--- a/.changeset/long-bears-fetch.md
+++ b/.changeset/long-bears-fetch.md
@@ -2,4 +2,4 @@
 "ts-essentials": patch
 ---
 
-Allow objects to be used in `StrictOmit`
+Allow only objects to be used in `StrictOmit`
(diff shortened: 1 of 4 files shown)


krzkaczor (approved):

ifeltsweet:
This for some reason breaks intersection types in some cases:

```
type A = { foo: any };
type StrictA<T extends A> = StrictOmit<T, 'foo'> & { foo: string };

function makeStrict<T extends A>(thing: T): StrictA<T> {
  return thing; // Type 'A' is not assignable to type 'StrictOmit<T, "foo">'
}
```

The error here is strange, for some reason, it ignores the intersection completely.

I'm not entirely sure if my example is 100% valid, but this pattern works with normal `Omit` so it's a little strange why it does not work with `StrictOmit`. It also works fine with `Merge`.

Beraliv:
> This for some reason breaks intersection types in some cases:
> 
> ```ts
> type A = { foo: any };
> type StrictA<T extends A> = StrictOmit<T, 'foo'> & { foo: string };
> 
> function makeStrict<T extends A>(thing: T): StrictA<T> {
>   return thing; // Type 'A' is not assignable to type 'StrictOmit<T, "foo">'
> }
> ```
> 
> The error here is strange, for some reason, it ignores the intersection completely.
> 
> I'm not entirely sure if my example is 100% valid, but this pattern works with normal `Omit` so it's a little strange why it does not work with `StrictOmit`. It also works fine with `Merge`.

Thanks for raising the problem!

I will check it later

Beraliv:
Sorry for postponing the answer, I will check it probably on Wednesday

PR #246, opened 2021-08-28
Add isExact with basic tests

krzkaczor:
Closes: https://github.com/krzkaczor/ts-essentials/issues/222

### DIFF
--- a/lib/functions.ts
+++ b/lib/functions.ts
@@ -14,6 +14,6 @@ export function assert(condition: any, msg: string = "no additional info provide
 
 export function noop(..._args: unknown[]): void {}
 
-export function isExact<T, EXACT_SHAPE>(x: Exact<T, EXACT_SHAPE>): EXACT_SHAPE {
+export const isExact = <ExpectedShape>() => <ActualShape>(x: Exact<ActualShape, ExpectedShape>): ExpectedShape => {
   return x;
-}
+};
--- a/lib/functions.ts
+++ b/lib/functions.ts
@@ -1,3 +1,5 @@
+import { Exact } from "./types";
+
 export class UnreachableCaseError extends Error {
   constructor(value: never) {
     super(`Unreachable case: ${value}`);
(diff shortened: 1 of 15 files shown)


test/index.ts
```diff
@@ -1201,3 +1202,26 @@ function testIsTuple() {
...
+  isExact<typeof abc, BC>(abc);
+
+  // BC is exactly BC
+  isExact<typeof bc, BC>(bc);
+  // @ts-expect-error BC2 is not exactly BC (C has different type)
+  isExact<typeof bc2, BC>(bc2);
+
+  // @ts-expect-error C is not exactly BC (missing property B)
+  isExact<typeof c, BC>(c);
+}
```
Beraliv, line 1227: wdyt about test cases with `as const`?

example: https://github.com/krzkaczor/ts-essentials/pull/247

lib/functions.ts
```diff
@@ -11,3 +13,7 @@ export function assert(condition: any, msg: string = "no additional info provide
 }
 
 export function noop(..._args: unknown[]): void {}
+
+export function isExact<T, EXACT_SHAPE>(x: Exact<T, EXACT_SHAPE>): EXACT_SHAPE {
+  return x;
+}
```
Beraliv, line 19: I found one workaround which help get rid of `typeof obj`: we create curried version of the function, but use only the first generic type (second will be inferred automatically) – https://github.com/krzkaczor/ts-essentials/pull/248

If I find something else, I will let you know

Feel free to merge or to close it

test/index.ts
Beraliv: Done here ✅ thanks for merging

lib/functions.ts
Beraliv: Merged too ✅

Beraliv (approved):
LGTM 🚀

Beraliv:
@krzkaczor LGTM, you can merge it 🚀

PR #248, opened 2021-08-28
suggestion: 🚀 isExact approach 2

Beraliv:
```ts
// the implementation
export const isExact = <U>() => <T>(object: Exact<T, U>) => {
  return object;
};

// the usage
isExact<BC>()(abc);

// or possible to reuse it later (which looks nicer)
const isBC = isExact<BC>();
isBC(abc);
```

I know extra `()` is kind of annoying, but imho less annoying than `typeof obj`

### DIFF
--- a/lib/functions.ts
+++ b/lib/functions.ts
@@ -14,6 +14,6 @@ export function assert(condition: any, msg: string = "no additional info provide
 
 export function noop(..._args: unknown[]): void {}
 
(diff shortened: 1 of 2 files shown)


krzkaczor:
This is perfect! Nice work! Will you fix the merge conflict?

Beraliv:
> This is perfect! Nice work! Will you fix the merge conflict?

Yeah, I will prepare the fix by tomorrow evening

Issue #257, opened 2021-09-20
Compilation Error (Merge) with TS 4.0.2 and ts-essentials 8.1.0


panewman:
I have a compilation error with Typescript 4.0.2 and ts-essentials 8.1.0
```
ERROR in node_modules/ts-essentials/dist/types.d.ts:315:14 - error TS2456: Type alias '_MergeN' circularly references itself.

315 declare type _MergeN<T extends readonly any[], Result> = T extends readonly [infer Head, ...(infer Tail)]
node_modules/ts-essentials/dist/types.d.ts:315:90 - error TS2574: A rest element type must be an array type.

315 declare type _MergeN<T extends readonly any[], Result> = T extends readonly [infer Head, ...(infer Tail)]
node_modules/ts-essentials/dist/types.d.ts:316:5 - error TS2315: Type '_MergeN' is not generic.

316   ? _MergeN<Tail, Merge<Result, Head>>
node_modules/ts-essentials/dist/types.d.ts:319:56 - error TS2315: Type '_MergeN' is not generic.

319 export declare type MergeN<T extends readonly any[]> = _MergeN<T, {}>;
```

Beraliv:
@panewman thank you for the issue!

Let me check it in the evening and I come back to you!

Beraliv:
@panewman thank you for reporting the issue!

You're absolutely right, recursive conditional types were introduced in [TypeScript@4.1](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-1.html#recursive-conditional-types) release

So we need to update the dependency and use >=4.1 instead of >=4.0

@krzkaczor do you think it makes sense?

Beraliv:
@panewman I will release a new version today

If you want to still use Typescript 4.0.2, please roll back to [v7.0.3](https://github.com/krzkaczor/ts-essentials/releases/tag/v7.0.3)

panewman:
Yes, I did that.

PR #258, opened 2021-09-26
fix: 🧪 Use TypeScript@^4.1.0 because of introduced recursive conditional types

Beraliv:
Related to #257

### DIFF
--- a/.changeset/good-knives-jam.md
+++ b/.changeset/good-knives-jam.md
@@ -1,0 +1,5 @@
+---
+"ts-essentials": patch
+---
+
+Use TypeScript@^4.1.0 because of introduced recursive conditional types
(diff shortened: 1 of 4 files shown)


.changeset/good-knives-jam.md
```diff
@@ -0,0 +1,5 @@
+---
+"ts-essentials": patch
+---
+
+Use TypeScript@^4.1.0 because of introduced recursive conditional types
```
Beraliv, line 5: maybe it makes sense to make it major change as we did when updated typescript peer dependency in [v8.0.0](https://github.com/krzkaczor/ts-essentials/releases/tag/v8.0.0)

krzkaczor: Yeah, i think it has to be major change as it can break deps.

Beraliv: Fixed ✅

PR #262, opened 2021-10-03
feat: ➕ add HomogeneousArray

Beraliv:


krzkaczor:
Could you expand on the rationale of this type?

From example:

```
type Arr = HomogeneousArray<string | number>;
```

I would much prefer to simply write:

```typescript
type Arr = string[] | number[];
```

This doesn't require knowledge of any helpers and it's self-contained.

Beraliv:
> I would much prefer to simply write:
> 
> ```typescript
> type Arr = string[] | number[];
> ```

@krzkaczor that's true

I just thought it can be helpful in some cases
However I would rarely use it myself

Beraliv:
Also I thought it can be also useful to transform:

```
type Source = (TypeA | TypeB | ... | TypeZ)[]
// transformation 💫
type Target = TypeA[] | TypeB[] | ... | TypeZ[]
```

Sometimes that's what you need

But again not sure about the use cases in production and real application, so for now we can close it actually

PR #267, opened 2021-11-07
build: yarn lockfile 2

Beraliv:
Migration to Yarn 3 – https://yarnpkg.com/getting-started/migration#step-by-step

Use node>=12.0.0 because of workers
- https://stackoverflow.com/questions/62280966/cannot-find-module-worker-threads-error-in-production
- Error looked like that: https://github.com/krzkaczor/ts-essentials/runs/4130574086?check_suite_focus=true


.github/workflows/ci.yml
```diff
@@ -28,7 +28,7 @@ jobs:
           key: ${{ runner.OS }}-${{ matrix.node }}-yarn-cache-${{ hashFiles('yarn.lock') }}
           restore-keys: ${{ runner.OS }}-${{ matrix.node }}-yarn-cache-
 
-      - run: yarn --no-progress --non-interactive --frozen-lockfile
+      - run: yarn install --immutable --immutable-cache --check-cache
```
Beraliv, line 31: Link to the change – https://yarnpkg.com/cli/install#examples:

> Validate a project when using Zero-Installs (slightly safer if you accept external PRs) :
```bash
yarn install --immutable --immutable-cache --check-cache
```

.gitignore
```diff
@@ -1,2 +1,4 @@
 node_modules
 dist
+.yarn/cache
```
Beraliv, line 3: I put them here because of – https://yarnpkg.com/getting-started/qa#which-files-should-be-gitignored

.github/workflows/ci.yml
Beraliv: Removed cache flags because of this weird error – https://github.com/krzkaczor/ts-essentials/runs/4130609656?check_suite_focus=true

package.json
```diff
@@ -32,6 +32,9 @@
   "peerDependencies": {
     "typescript": ">=4.1.0"
   },
+  "engines": {
```
krzkaczor, line 35: can we add `.nvmrc` file too?

krzkaczor (commented):
Looking good! TBH I've never used yarn 2.x but I'm happy to give it a chance. Especially that it's not that invasive.

package.json
Beraliv: yeah

krzkaczor:
btw. is dumping yarn as `.cjs` a common practice? I would much prefer to have a minimal script that would download it from the official distro site + verify signatures + cache locally.

Beraliv:
I'm thinking to switch to `pnpm` instead

krzkaczor:
@Beraliv i got a chance to check `pnpm` and I love it. Totally agree to migrate.

Beraliv:
> @Beraliv i got a chance to check `pnpm` and I love it. Totally agree to migrate. 

Me too, and I really loved it as well

I think it's worth to move to it

PR #270, opened 2021-11-08
feat: 🔄 update DeepOmit

Beraliv:
- [x] Add generic constrain on Filter
- [x] Cover everything with tests
- [x] Iterate over the keys once

### DIFF
--- a/lib/types.ts
+++ b/lib/types.ts
@@ -226,47 +226,46 @@ export type RequiredKeys<T> = T extends unknown ? Exclude<keyof T, OptionalKeys<
 export type PickKeys<T, P> = Exclude<keyof PickProperties<T, P>, undefined>;
 
 /** Recursively omit deep properties */
-// explicitly mentioning optional properties, to work around TS making them required
-// see https://github.com/krzkaczor/ts-essentials/issues/118
-export type DeepOmit<T extends DeepOmitModify<Filter>, Filter> = T extends Builtin
+export type DeepOmit<T, Filter extends DeepModify<T>> = T extends Builtin
   ? T
   : T extends Map<infer KeyType, infer ValueType>
-  ? ValueType extends DeepOmitModify<Filter>
-    ? Map<KeyType, DeepOmit<ValueType, Filter>>
+  ? Filter extends DeepModify<ValueType>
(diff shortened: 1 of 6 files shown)


lib/types.ts
```diff
@@ -226,64 +228,56 @@ export type RequiredKeys<T> = T extends unknown ? Exclude<keyof T, OptionalKeys<
...
-        },
-        never
-      >;
+          : K
+        : K]: K extends keyof Filter
+        ? Filter[K & keyof Filter] extends DeepModify<T[K]>
+          ? DeepOmit<T[K], Filter[K & keyof Filter]>
+          : T[K]
+        : T[K];
+    }
```
Beraliv, line 276: Now it returns:

```ts
let obj10: DeepOmit<ComplexNestedRequired, { nested: { date: true; array: true } }>;
```

And I see this for `obj10`:

```ts
let obj10: {
    simple: number;
    nested: {
        func: () => string;
        tuple: [string, number, {
            good: boolean;
        }];
        set: Set<{
            name: string;
        }>;
        map: Map<string, {
            name: string;
        }>;
        promise: Promise<{
            ...;
        }>;
    };
}
```

lib/types.ts
```diff
@@ -226,64 +228,56 @@ export type RequiredKeys<T> = T extends unknown ? Exclude<keyof T, OptionalKeys<
...
-    > & { [K in Exclude<RequiredKeys<T>, keyof Filter>]: T[K] } & OmitProperties<
-        {
-          [K in Extract<RequiredKeys<T>, keyof Filter>]: Filter[K] extends true
-            ? never
-            : T[K] extends DeepOmitModify<Filter[K]>
-            ? DeepOmit<T[K], Filter[K]>
-            : T[K];
-        },
-        never
-      >;
```
Beraliv, line 283: Here there was:

```ts
let obj10: {} & OmitProperties<{}, never> & {
    simple: number;
} & OmitProperties<{
    nested: {} & OmitProperties<{}, never> & {
        map: Map<string, {
            name: string;
        }>;
        func: () => string;
        tuple: [...];
        set: Set<...>;
        promise: Promise<...>;
    } & OmitProperties<...>;
}, never>
```

krzkaczor (approved):

Issue #272, opened 2021-11-23
Fix DeepPick & DeepOmit for set, map, array and promise


Beraliv:
Currently it includes `DeepPick` by mistake – https://github.com/krzkaczor/ts-essentials/blob/master/lib/types.ts#L231

Beraliv:
Also found here that we cannot work with types other than objects at the moment

Beraliv:
I fixed an issue with both `DeepPick` and `DeepOmit` and fixed it in https://github.com/krzkaczor/ts-essentials/pull/273

I will also include it in https://github.com/krzkaczor/ts-essentials/pull/264

Beraliv:
Sorry, had problems with CI, release under https://github.com/krzkaczor/ts-essentials/pull/280

PR #273, opened 2021-12-06
fix: 🧪 DeepPick & DeepOmit for set, map, promise and array

Beraliv:
Relates to #272

### DIFF
--- a/lib/types.ts
+++ b/lib/types.ts
@@ -231,36 +231,36 @@ export type PickKeys<T, P> = Exclude<keyof PickProperties<T, P>, undefined>;
 export type DeepOmit<T, Filter extends DeepModify<T>> = T extends Builtin
   ? T
   : T extends Map<infer KeyType, infer ValueType>
-  ? Filter extends DeepModify<ValueType>
-    ? Map<KeyType, DeepPick<ValueType, Filter>>
+  ? Filter extends Map<KeyType, infer FilterValueType>
+    ? Map<KeyType, DeepOmit<ValueType, FilterValueType>>
     : T
   : T extends ReadonlyMap<infer KeyType, infer ValueType>
-  ? Filter extends DeepModify<ValueType>
-    ? ReadonlyMap<KeyType, DeepPick<ValueType, Filter>>
+  ? Filter extends ReadonlyMap<KeyType, infer FilterValueType>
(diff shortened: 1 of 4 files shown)


lib/types.ts
```diff
@@ -330,13 +330,13 @@ type DeepModify<T> =
...
-  | WeakSet<DeepModify<T>>
-  | Map<any, DeepModify<T>>
-  | WeakMap<any, DeepModify<T>>;
+  | (T extends Array<infer E> ? Array<DeepModify<E>> : never)
+  | (T extends Promise<infer E> ? Promise<DeepModify<E>> : never)
+  | (T extends Set<infer E> ? Set<DeepModify<E>> : never)
+  | (T extends ReadonlySet<infer E> ? ReadonlySet<DeepModify<E>> : never)
+  | (T extends WeakSet<infer E> ? WeakSet<DeepModify<E>> : never)
+  | (T extends Map<any, infer E> ? Map<any, DeepModify<E>> : never)
+  | (T extends WeakMap<any, infer E> ? WeakMap<any, DeepModify<E>> : never);
```
bbarry, line 339: Should these last 2 be

```typescript
  | (T extends Map<infer K, infer E> ? Map<K, DeepModify<E>> : never)
  | (T extends WeakMap<infer K, infer E> ? WeakMap<K, DeepModify<E>> : never);
```

I kinda like this union type style vs the deep nested conditional style of DeepPick/DeepOmit.


Does this change belong here at all? Are there any tests for this change?

Beraliv: That's a good catch! 👀

I think it doesn't really change anything because we are focusing more on values rather than keys

Anyway I need to add tests for at least all maps

Beraliv: However something is covered here – https://github.com/krzkaczor/ts-essentials/pull/273/files#diff-8c10bb7fdebd2b6fa3438e87e51411024b7af2ccb5abd52dd83e52cadae49a45R618

See that `number` cannot be used for `Maps` because it was declared `string`

krzkaczor (approved):

lib/types.ts
Beraliv: I covered it ✅

Now it should be strictly equal to the value

Issue #278, opened 2021-12-20
ts-essentials breaks with keyofStringsOnly typescript compiler option


MarcCelani-at:
This issue was first reported in https://github.com/krzkaczor/ts-essentials/issues/46, and the user just removed the compiler option in their project, which does not work for us.

This is the error when typechecking our project with skipLibCheck=false and keyofStringsOnly=true

```
node_modules/ts-essentials/dist/types.d.ts:5:13 - error TS2322: Type 'K' is not assignable to type 'string'.
  Type 'string | number' is not assignable to type 'string'.
    Type 'number' is not assignable to type 'string'.

      [key in K]: T;
```

MarcCelani-at:
The error appears to go away if I patch Dictionary to be...

```
export declare type Dictionary<T, K extends keyof any = string> = {
    [key in K]: T;
};
```

Of course, if you do not want to allow symbols, you may be able to do something like keyof any & (string | number)?

Beraliv:
Hey @MarcCelani-at !

Thank you so much for the report!

I've reproduced it here – https://tsplay.dev/wEGEZm

To be able to fix it, we should define out `PropertyKey` version – https://devblogs.microsoft.com/typescript/announcing-typescript-2-9-2/#:~:text=type%20KeyofBase%20%3D%20keyof%20any%3B

It will look like:

```ts
type KeyofBase = keyof any;
```

It equals to `string` when you enable the option – https://tsplay.dev/mZb7Pm
Otherwise, it will return `string | number | symbol` – https://tsplay.dev/Ndd7kN

I will fix it in the nearest release

Beraliv:
@MarcCelani-at can you please check the changes from https://github.com/krzkaczor/ts-essentials/pull/279? Does it fix your problem, right?

Beraliv:
Included it in the next release – https://github.com/krzkaczor/ts-essentials/pull/280

PR #279, opened 2021-12-20
fix: 🧪 tolerate keyofStringsOnly option in TypeScript

Beraliv:
Relates to #278

### DIFF
--- a/.changeset/neat-impalas-beam.md
+++ b/.changeset/neat-impalas-beam.md
@@ -1,0 +1,5 @@
+---
+"ts-essentials": patch
+---
+
+Replace `PropertyKey` with `KeyofBase` to tolerate `keyofStringsOnly` option in TypeScript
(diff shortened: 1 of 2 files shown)


MarcCelani-at:
It'll take a bit to test these changes on my project since we are on a very old version of ts-essentials (I don't have AnyRecord or SafeDictionary)

```
(env-12.20.1) ➜  hyperbase git:(marccelani-backboneRouterRefactor) yarn why ts-essentials
├─ use-composed-ref@npm:1.0.0
│  └─ ts-essentials@npm:2.0.12 (via npm:^2.0.3)
│
└─ use-composed-ref@npm:1.0.0 [d2064]
   └─ ts-essentials@npm:2.0.12 (via npm:^2.0.3)
```

But what you did seems pretty similar to how I got this working.

MarcCelani-at:
I've confirmed that this patch will work for our project. I'd love if you could land this and bump the version.

Beraliv:
> I've confirmed that this patch will work for our project. I'd love if you could land this and bump the version.

Perfect, thank you for your check!

I will try to include it in the next release

krzkaczor (approved):

PR #304, opened 2022-02-28
docs: type error in DeepPick/DeepOmit example

liaoyinglong:
miss `[]` in DeepPick/DeepOmit example

before
image

this pr
image


Beraliv:
@liaoyinglong thank you for the fix!

Beraliv (approved):

PR #305, opened 2022-03-11
add PredicateType and PredicateFunction interfaces and full set of tests

### LINKED ISSUE #253: PredicateType: pull the predicate off of a type guard or other function with predicate attached
I've often had a need to chain type guards in a manner than continues to narrow in on the type, for usability purposes. Unfortunately, TS doesn't provide a type guard's predicate details to allow such chaining (context is lost once outside the guarding function). Finally figured it out after months of playing around and thought I'd offer it up for consideration in this lib.

```ts
/**
 * For use to extract the Predicate that has been set on a function
 */
export type PredicateType<
  T extends (...x: any[]) => boolean
> = T extends (target: any, ...rest: any[]) => target is infer P ? P : never
```

Contrived example use case:
```ts
const isArrayOf = <
  T extends (...x: any[]) => boolean
>(thing: unknown, validator: T): thing is Array<PredicateType<T>> => (
  Array.isArray(thing)
  && thing.every(validator)
)

const isMessage = (thing: unknown): thing is { type: "message" } => (
  typeof thing === "object"
  && thing !== null
  && 'type' in thing
  && thing['type'] === "message"
)

const thing = []
if(isArrayOf(thing, isMessage)) { /* do something */ }
```

Below is some other examples that could use this and the outcomes from using the type.

```ts
// no type predicate set: never
const never = (thing: any) => true
type neverType = PredicateType<typeof never>

// return basic predicate
const basic = (thing: unknown): thing is boolean => typeof thing === 'boolean'
type testtype2 = PredicateType<typeof basic>

// returns predicate even if more arguments are supplied
const extraArgs = (thing: unknown, _other: any): thing is boolean => typeof thing === 'boolean'
type extraArgsType = PredicateType<typeof extraArgs>

// returns a complex predicate
const complexPredicate = (thing: { type: string }): thing is typeof thing & { payload: unknown } => 'payload' in thing
type complexPredicateType = PredicateType<typeof complexPredicate>

// if predicate is not first: never (i'd rather this error, but its technically true as written 🤷‍♂️; see below)
const lastArgPredicate = (thing: unknown, other: unknown): other is boolean => typeof other === 'boolean'
type lastArgPredicateType = PredicateType<typeof lastArgPredicate>

// type error when used with asserts
const testAssert = (thing: unknown): asserts thing is false => {
  if(typeof thing === 'boolean' && thing === false) {
    throw new Error('why you giving me falsy things?')
  }
}
// @ts-expect-error Type '(thing: unknown) => asserts thing is false' does not satisfy the constraint '(...x: any[]) => boolean'.
type testAssertType = PredicateType<typeof testAssert>
```

As for "if predicate is not first: never", I've never had a use case where I wanted to target outside the first argument, but this could be expanded as such (or with overloads) should you want that use case... like perhaps following a functional mentality and having target supplied "last":

```ts
export type PredicateType<T extends (...x: any[]) => boolean> =
  T extends (target: any, ...rest: any[]) => target is infer P ? P
  : T extends (arg1:any, target: any, ...rest: any[]) => target is infer P ? P
  : T extends (arg1:any, arg2: any, target: any, ...rest: any[]) => target is infer P ? P
  : never
```

A similar construct might work for `asserts` as well, but honestly I'm not sure how useful that'd be at all.

edit: formatting

kevinpeno:
Fixes #253

1. Adds type called `PredicateType` to support snagging predicates off of guarded functions
2. Adds utility type called `PredicateFunction` to let folks duplicate the interface for things like the `isArrayOf` [example in the tests](https://github.com/ts-essentials/ts-essentials/blob/3fddc463985967e6084a9a7e0d5009b15d6ef793/test/predicate-type.ts#L36). I left it undocumented on the readme because this doesn't seem like a normal thing to need, but it is easily discovered

### DIFF
--- a/README.md
+++ b/README.md
@@ -1070,6 +1070,24 @@ assert(anything instanceof String, "anything has to be a string!");
 // from now on `anything` is string
 ```
 
+### PredicateType
+_keywords: narrow, guard, validate_
+
+Works just like [`ReturnType`](https://www.typescriptlang.org/docs/handbook/utility-types.html#returntypetype) but will return the [predicate](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates) associated with the function instead. This is particularly useful if you need to chain guards to narrow broader types.
+
+```typescript
+// Without PredicateType you can never use a set of functions like this together; how can you resolve ???
+// You would need a specific instance of isArrayOf for each type you want to narrow
+const isArrayOf = (thing: unknown, validator: (...x: any[]) => boolean): thing is ???[] => {
+  return Array.isArray(thing) && thing.every(validator)
+}
+
+// With PredicateType you can pull the predicate of the validator into the higher level guard
+const isArrayOf = <T extends (...x: any[]) => boolean>(thing: unknown, validator: T): thing is Array<PredicateType<T>> => {
+  return Array.isArray(thing) && thing.every(validator)
+}
+```
+
 ### Exact
 
 _keywords: same, equals, equality_
--- a/lib/types.ts
+++ b/lib/types.ts
@@ -475,3 +475,15 @@ export type Tail<T extends AnyArray> = T["length"] extends 0
   : never;
 
 export type Exact<T, SHAPE> = T extends SHAPE ? (Exclude<keyof T, keyof SHAPE> extends never ? T : never) : never;
+
+/**
+ * Basic interface for guarded functions that use [predicates](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates)
+ */
+export type PredicateFunction = (x: any, ..._z: any[]) => x is any;
+
+/**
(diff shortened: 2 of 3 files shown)


kevinpeno:
@Beraliv reminder that this is up.

Beraliv:
> @Beraliv reminder that this is up.

Hey!

Sorry for the late response and thank you for your PR!

I moved to another country recently and still try to manage my schedule 😅

I will try to have a look at it closer to the night today, sorry for the inconvenience in advance!

Beraliv (approved):
thank you for proper testing, LGTM

let's merge it!

kevinpeno:
@Beraliv I don't have a merge option. Is there a bot command I need to run?

Beraliv:
> @Beraliv I don't have a merge option. Is there a bot command I need to run?

I will do it for you later, sorry for the inconvenience!

Issue #307, opened 2022-04-03
NonEmptyArray


yss14:
Similar to `NonEmptyObject`, `NonEmptyArray` can also be a useful type essential in typescript codebases, specifically when working with [rest parameters](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/rest_parameters). 

Implementation:
```typescript
type NonEmptyArray<T> = [T, ...T[]]

const okay: NonEmptyArray<number> = [1, 2];
const alsoOkay: NonEmptyArray<number> = [1];
const error: NonEmptyArray<number> = []; // Type '[]' is not assignable to type 'NonEmpty<number>'. Source has 0 element(s) but target requires 1.
```

Usecase Example:
```typescript
// Implementing some custom express middleware which requires at least one rest parameter
type RESTMethod<Request> = {
	(
		path: string,
		check: TypedRequestHandler<Request>,
		...middlewares: NonEmptyArray<TypedRequestHandler<Request>>
	): TypedRouter<Request>
}
```

I would be happy to submit a PR for the `NonEmptyArray` type.

Beraliv:
Hey @yss14!

Sorry for the late response!

Thank you for the suggestion! Sounds like a great feature!

If you have time for PR, could you please implement it?

I'm still struggling with my schedule after the relocation to another country and try to get back to normal one by the end of April.

Sorry for the inconvenience in advance and thank you for understanding!

yss14:
@Beraliv Thanks for your response and no worries, 7 days is not a late response in open source world :)

I just submitted a PR for that feature: https://github.com/ts-essentials/ts-essentials/pull/308

Beraliv:
It will be released here – https://github.com/ts-essentials/ts-essentials/pull/302

I will let you know when it's done and will close this issue

Thank you for your patience!

Beraliv:
It's now available in 9.2.0 – https://github.com/ts-essentials/ts-essentials/releases/tag/v9.2.0

PR #308, opened 2022-04-10
feat: add NonEmptyArray type

yss14:
**Changes:**
* add new `NonEmptyArray` type
* add missing readme menu entry for `NonEmptyObject` type

Resolves https://github.com/ts-essentials/ts-essentials/issues/307

### DIFF
--- a/README.md
+++ b/README.md
@@ -670,7 +670,7 @@ Useful for accepting only arrays containing at least one element.
 
 ```typescript
 type NonEmptyRestParameters<T> = NonEmptyArray<T>;
-// Result: [number, ...number[]]
+// Result: [T, ...T[]]
 
 // declare function expression type accepting some rest parameters, but at least one element for the rest parameters is required
 type FunctionAcceptingRestParameters = (someString: string, ...args: NonEmptyRestParameters<number>) => void;
--- a/README.md
+++ b/README.md
@@ -669,11 +669,8 @@ type EmptyObject = NumberDictionary<{}>;
 Useful for accepting only arrays containing at least one element.
 
 ```typescript
-type NonEmptyRestParameters<T> = NonEmptyArray<T>;
-// Result: [T, ...T[]]
-
 // declare function expression type accepting some rest parameters, but at least one element for the rest parameters is required
-type FunctionAcceptingRestParameters = (someString: string, ...args: NonEmptyRestParameters<number>) => void;
+type FunctionAcceptingRestParameters = (someString: string, ...args: NonEmptyArray<number>) => void;
 
 // declare some non-empty array variables
 const okay: NonEmptyArray<number> = [1, 2];
--- a/README.md
+++ b/README.md
@@ -669,7 +669,7 @@ type EmptyObject = NumberDictionary<{}>;
 Useful for accepting only arrays containing at least one element.
 
 ```typescript
-type NonEmptyRestParameters = NonEmptyArray<number>;
+type NonEmptyRestParameters<T> = NonEmptyArray<T>;
 // Result: [number, ...number[]]
 
 // declare function expression type accepting some rest parameters, but at least one element for the rest parameters is required
@@ -678,7 +678,8 @@ type FunctionAcceptingRestParameters = (someString: string, ...args: NonEmptyRes
 // declare some non-empty array variables
 const okay: NonEmptyArray<number> = [1, 2];
 const alsoOkay: NonEmptyArray<number> = [1];
-const error: NonEmptyArray<number> = []; // Type '[]' is not assignable to type 'NonEmptyArray<number>'. Source has 0 element(s) but target requires 1.
+// @ts-expect-error: Type '[]' is not assignable to type 'NonEmptyArray<number>'. Source has 0 element(s) but target requires 1.
+const error: NonEmptyArray<number> = [];
 ```
 
 ### Merge
--- a/README.md
+++ b/README.md
@@ -62,6 +62,8 @@ If you use any [functions](https://github.com/krzkaczor/ts-essentials/blob/maste
   - [OmitProperties](#OmitProperties)
   - [PickProperties](#PickProperties)
   - [NonNever](#NonNever)
+  - [NonEmptyObject](#NonEmptyObject)
+  - [NonEmptyArray](#NonEmptyArray)
   - [Merge](#Merge)
   - [MergeN](#MergeN)
   - [MarkRequired](#MarkRequired)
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -662,6 +664,23 @@ type SomeObject = NumberDictionary<{ a: number; b: string }>;
...
+### NonEmptyArray
+
+Useful for accepting only arrays containing at least one element.
+
+```typescript
+type NonEmptyRestParameters = NonEmptyArray<number>;
+// Result: [number, ...number[]]
+
+// declare function expression type accepting some rest parameters, but at least one element for the rest parameters is required
+type FunctionAcceptingRestParameters = (someString: string, ...args: NonEmptyRestParameters<number>) => void;
```
Beraliv, line 676: Would be cool to have this feature in TypeScript, but it's not implemented yet 😅

```suggestion
type FunctionAcceptingRestParameters = (someString: string, ...args: NonEmptyRestParameters) => void;
```

README.md
```diff
@@ -662,6 +664,23 @@ type SomeObject = NumberDictionary<{ a: number; b: string }>;
...
+
+// declare some non-empty array variables
+const okay: NonEmptyArray<number> = [1, 2];
+const alsoOkay: NonEmptyArray<number> = [1];
+const error: NonEmptyArray<number> = []; // Type '[]' is not assignable to type 'NonEmptyArray<number>'. Source has 0 element(s) but target requires 1.
```
Beraliv, line 681: Might be worth to have 2 lines instead of 1:

```suggestion
// @ts-expect-error: Type '[]' is not assignable to type 'NonEmptyArray<number>'. Source has 0 element(s) but target requires 1.
const error: NonEmptyArray<number> = [];
```

WDYT?

test/index.ts
```diff
@@ -832,6 +833,16 @@ function testNonEmptyObject() {
...
 }
 
+function testNonEmptyArray() {
+  type TestA<T> = Assert<IsExact<NonEmptyArray<T>, [T, ...T[]]>>;
+  type TestB<T> = AssertFalse<IsExact<NonEmptyArray<T>, []>>;
+
+  type TestEmptyArrayNotAssignable<T> = AssertFalse<Assignable<NonEmptyArray<T>, []>>;
+  type TestArrayOneElementAssignable<T> = Assert<Assignable<NonEmptyArray<T>, [T]>>;
+  type TestArrayTwoElementsAssignable<T> = Assert<Assignable<NonEmptyArray<T>, [T, T]>>;
+  type TestArrayMoreThanOneElementsAssignable<T> = Assert<Assignable<NonEmptyArray<T>, [T, ...T[]]>>;
```
Beraliv, line 843: I know that examples in this file are not well structured, but could you please change it to:

```ts
type cases = [
  Assert<IsExact<NonEmptyArray<T>, [T, ...T[]]>>,
  AssertFalse<IsExact<NonEmptyArray<T>, []>>,
  ....
];
```

I think it contains less types that we don't use anyway. And if you want to highlight it somehow, you can write comment right before the test case

WDYT?

Beraliv (changes requested):
Thank you for the help! LGTM! 👍

There are only several places that I would adjust a little bit before merging it.

README.md
yss14: Ah damn, I think my original idea was to make NonEmptyRestParameters generic :D

yss14:
> Thank you for the help! LGTM! 👍
> 
> There are only several places that I would adjust a little bit before merging it.

Thanks for your review and I totally agree with your comments and suggestions!
I pushed some new commits which implement the proposals.
Looking forward to a second review from your side :)

test/index.ts
```diff
@@ -832,6 +833,17 @@ function testNonEmptyObject() {
...
+function testNonEmptyArray() {
+  type Cases<T> = [
+    Assert<IsExact<NonEmptyArray<T>, [T, ...T[]]>>,
+    AssertFalse<IsExact<NonEmptyArray<T>, []>>,
+
+    AssertFalse<Assignable<NonEmptyArray<T>, []>>,
+    Assert<Assignable<NonEmptyArray<T>, [T]>>,
+    Assert<Assignable<NonEmptyArray<T>, [T, ...T[]]>>,
+  ];
+}
```
Beraliv, line 845: Thank you for the change and all in all, smart idea with generic type in `Cases<T>`!

README.md
```diff
@@ -662,6 +664,24 @@ type SomeObject = NumberDictionary<{ a: number; b: string }>;
...
+### NonEmptyArray
+
+Useful for accepting only arrays containing at least one element.
+
+```typescript
+type NonEmptyRestParameters<T> = NonEmptyArray<T>;
+// Result: [T, ...T[]]
+
+// declare function expression type accepting some rest parameters, but at least one element for the rest parameters is required
+type FunctionAcceptingRestParameters = (someString: string, ...args: NonEmptyRestParameters<number>) => void;
```
Beraliv, line 676: Not critical comment

You can use it here as is:

```suggestion
type FunctionAcceptingRestParameters = (someString: string, ...args: NonEmptyArray<number>) => void;
```

And then don't forget to remove previous lines

Beraliv (approved):
Thank you for the changes and all in all LGTM!

README.md
yss14: Makes sense to simplify the example here a little bit :)

yss14:
@Beraliv I pushed the changes. If it's all fine now, I guess you must merge the PR, since I'm lacking permissions to do so. But no hurries :)

psmolinsky:
Hi guys, btw wouldn't it be also useful to have a type predicate for the `NonEmptyArray` (or `NonEmptyObject` too)? Example:

```typescript
export function isNonEmpty<T>(array: T[] | null | undefined): array is NonEmptyArray<T>;
export function isNonEmpty<T>(array: readonly T[] | null | undefined): array is Readonly<NonEmptyArray<T>>;
export function isNonEmpty<T>(array: readonly T[] | null | undefined): boolean {
    return Array.isArray(array) && array.length > 0;
}
```

yss14:
@psmolinsky I would think that this might be out of scope since ts-essentials is all about typescript types.

psmolinsky:
One more suggestion: would it be useful to have a readonly version of `NonEmptyArray<T>`? My first guess was that `Readonly<NonEmptyArray<T>>` would do the job but it still has some properties writable e.g. `length`:

```typescript
const array: Readonly<NonEmptyArray<number>> = Object.freeze([1, 2] as const);

array.length = 10; // This compiles but fails in runtime.
```

Beraliv:
> One more suggestion: would it be useful to have a readonly version of `NonEmptyArray<T>`? My first guess was that `Readonly<NonEmptyArray<T>>` would do the job but it still has some properties writable e.g. `length`:
> 
> ```ts
> const array: Readonly<NonEmptyArray<number>> = Object.freeze([1, 2] as const);
> 
> array.length = 10; // This compiles but fails in runtime.
> ```

Hey @psmolinsky !

Thank you for the suggestion!

I think in your case you don't need to have `Readonly<NonEmptyArray<number>>`

If you're worried that you can miss empty array here somehow:

1. You can put your configuration (or array) into a separate file
2. Add a test for your file checking if it's empty or not
3. Use it in other files without checking (then `array.length = <number>` will be prohibited)

Playground – https://tsplay.dev/NDk91w

Did I understand you case correctly or there is something I missed here?

Beraliv:
@yss14 could you please also add changelog with `yarn changelog`?

The example is here: https://github.com/ts-essentials/ts-essentials/commit/6b27ee4b7849d2a2670d80000a928cd20e3981b7

I think it should be `minor` as it's adding a new feature

yss14:
@Beraliv Sure thing! I've added a changeset to reflect the minor change.

Beraliv:
Thank you for your changes, I will merge it now

Sorry for waiting, hope it won't block you anyhow

PR #321, opened 2022-07-07
refactor(tail): use variadic tuple type

anuraghazra:
Refactored the older function signature to variadic tuple type introduced in TS 4.0

### DIFF
--- a/lib/types.ts
+++ b/lib/types.ts
@@ -470,7 +470,7 @@ export type XOR<T, U> = T | U extends object ? (Without<T, U> & U) | (Without<U,
 
 /** Functional programming essentials */
 export type Head<T extends AnyArray> = T["length"] extends 0 ? never : T[0];
-export type Tail<T extends AnyArray> = T["length"] extends 0 ? never : T extends [any, ...infer Rest] ? Rest : never;
+export type Tail<T extends AnyArray> = T extends [any, ...infer Rest] ? Rest : never;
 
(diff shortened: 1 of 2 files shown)


lib/types.ts
```diff
@@ -470,11 +470,7 @@ export type XOR<T, U> = T | U extends object ? (Without<T, U> & U) | (Without<U,
 
 /** Functional programming essentials */
 export type Head<T extends AnyArray> = T["length"] extends 0 ? never : T[0];
-export type Tail<T extends AnyArray> = T["length"] extends 0
-  ? never
-  : ((...t: T) => void) extends (first: any, ...rest: infer Rest) => void
-  ? Rest
-  : never;
+export type Tail<T extends AnyArray> = T["length"] extends 0 ? never : T extends [any, ...infer Rest] ? Rest : never;
```
Beraliv, line 473: You can even remove first check as it doesn't match next conditional type and will end up being `never` anyway

```suggestion
export type Tail<T extends AnyArray> = T extends [any, ...infer Rest] ? Rest : never;
```

Beraliv (changes requested):
Hey!

Thank you so much for your contribution! Good job and thank you for your changes!

I addressed you with just one comment how it can be even shorter.

Also don't forget to add changelog with `yarn changelog`

lib/types.ts
anuraghazra: Aha! Good catch! Thanks.

Beraliv (approved):
LGTM, thank you for your time and the changes!

Beraliv:
I will merge it as soon as I fix `master` branch CI problems! I will keep you updated!

Issue #323, opened 2022-07-17
Issue after upgrading from 9.1.2 to 9.2.0.


mehimanshupatil:
I am using type guard for a function which works flawlessly in v9.1.2 but when I upgraded to v9.2.0 it started giving error.
`A type predicate's type must be assignable to its parameter's type.`

```
import { MarkRequired } from "ts-essentials";

export enum AttachmentType {
  "IMAGE" = "IMAGE",
  "FILE" = "FILE",
  "CODE" = "CODE"
}

type AttachmentToBeUploaded =
  | {
      attachmentType: AttachmentType.IMAGE;
      id: string;
      localImageFileUrl?: string;
      localFile?: File;
      attachmentTypeSpecificFields: {
        name: string;
        type: string;
      };
    }
  | {
      attachmentType: AttachmentType.FILE;
      id: string;
      localImageFileUrl?: string;
      localFile?: File;
      attachmentTypeSpecificFields: {
        name: string;
      };
    }
  | {
      attachmentType: AttachmentType.CODE;
      id: string;
      attachmentTypeSpecificFields: {
        language: string;
      };
    };

export default function App() {
  const isAttachmentUploadable = (
    upload: AttachmentToBeUploaded
  ): upload is MarkRequired<
    Extract<
      AttachmentToBeUploaded,
      | { attachmentType: AttachmentType.IMAGE }
      | { attachmentType: AttachmentType.FILE }
    >,
    "localFile"
  > =>
    (upload.attachmentType === AttachmentType.FILE ||
      upload.attachmentType === AttachmentType.IMAGE) &&
    Boolean(upload.localFile);

  return (
    <></>
  );
}

```

reproducible codesandbox => [https://codesandbox.io/s/type-guard-error-plpnsu](https://codesandbox.io/s/type-guard-error-plpnsu)
change ts-essentials version in dependencies

Beraliv:
Hey @mehimanshupatil!

Thank you for your report!

I checked codesandbox and think the type error makes sense now because of the changes within 9.2.0

The change was introduced here – https://github.com/ts-essentials/ts-essentials/pull/295:

```diff
- export type MarkRequired<T, RK extends keyof T> = Exclude<T, RK> & Required<Pick<T, RK>>;
+ export type MarkRequired<T, RK extends keyof T> = Omit<T, RK> & Required<Pick<T, RK>>;
```

[...]

PR #324, opened 2022-07-18
feat: ➕ add union support for Mark* methods

Beraliv:
## What

Add union support to all Mark* methods: `MarkRequired`, `MarkOptional`, `MarkReadonly` and `MarkWritable`

## Why

Fix introduced bug – https://github.com/ts-essentials/ts-essentials/issues/323

Codesandbox fix – https://codesandbox.io/s/type-guard-error-forked-dyllt8?file=/src/App.tsx

### DIFF
--- a/.changeset/nasty-plants-check.md
+++ b/.changeset/nasty-plants-check.md
@@ -1,0 +1,5 @@
+---
+"ts-essentials": patch
+---
+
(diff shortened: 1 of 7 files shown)


Beraliv:
> @Beraliv one question on what `T extends T` does, isn't it will be always true?

It's a way to use distributive conditional type – https://www.typescriptlang.org/docs/handbook/2/conditional-types.html#distributive-conditional-types

You can write it different ways, that's just a short one – https://tsplay.dev/NV4vvW

Issue #315, opened 2022-06-22
Deep analogue of MarkRequired with the skeleton of DeepModify


xenoterracide:
I would like to be able to provide a structure to `DeepRequired` like DeepOmit, and I suspect this would be useful to all the Deep nested. The same structure as `DeepOmit` would work.

```ts
type TeacherSimple = DeepRequired<
  Teacher,
  {
    gender: never;
    students: {
      score: never;
    };
  }
>;
```

Beraliv:
Hey @xenoterracide!

Thank you for the suggestion! Yeah, it's a good feature that can be implemented

Beraliv:
I started working on it here – https://github.com/ts-essentials/ts-essentials/pull/329

It will take time, so it will be possibly available in the release after the next one

Beraliv:
I think I will create a separate `DeepMarkRequired` to not change the simplicity of `DeepRequired`

jrnail23:
I'd love this too.

Beraliv:
@jrnail23 hey! currently, it's a part of the next v10.1 release, but I can have a look at it now. I will come back to you when I have an implementation.

Beraliv:
After reflecting on the proposal, I've realised that the `Filter` object-like structure (similar to `DeepOmit` and `DeepPick`) would have an obvious disadvantage: it is unclear what properties have to be required: leaves only or all keys included in the filter?

A simple example from the issue proposal:

```ts
// There is a teacher with all required properties
type Teacher = {
  name: string;
  gender: "M" | "F"; // etc
  address: {
    postcode: string;
    city: string;
    street: string;
    apt: string;
  };
};

// Developers have to validate the object from the backend, therefore all properties are marked as optional
type MaybeTeacher = DeepPartial<Teacher>;

// But some of the properties have to be required for business logic, e.g. a city where a teacher lives
type TeacherWithAddress = DeepMarkRequired<
  MaybeTeacher,
  {
    address: {
      city: never;
    };
  }
>;
```

What is expected to get in `TeacherWithAddress`?

```ts
// 1. Only city is a required property while address will be optional
// {
//   name?: string;
//   gender?: "M" | "F"; // etc
//   address?: { // address is NOT required 🔴
//     postcode?: string;
//     city: string; // city is required 🟢
//     street?: string;
//     apt?: string;
//   };
// }

// 2. Both address and city will be required
// {
//   name?: string;
//   gender?: "M" | "F"; // etc
[...]

PR #329, opened 2022-08-19
feat: ➕ Add filter to DeepRequired

Beraliv:
## What

Add `Filter` to `DeepRequired`

## Why

To be able to mark readonly fields in deep structures

Related – #315


Beraliv:
I'm planning to close it as want to create a separate utility type for addressing the structure, it will be prefixed with `MarkDeep`

PR #334, opened 2022-10-28
Improve the Output range and cover edge cases for `ValueOf<T>` type

### LINKED ISSUE #333: Improve the Range for Type Helpers like `ValueOf`
### Definition

```ts
 type ValueOf<T> = T[keyof T] 
```
As per my understanding, the Domain of the `valueOf` type helper since being generic can be any Type Parameter (T)
Hence the Range for this Mapping should also be defined to cover it.

As you can see for example 2 lists out all the properties of the string that might not be sound given the Domain as a string.

```ts 
const obj = {
  id: "123e4567-e89b-12d3-a456-426655440000",
  name: "Test object",
  timestamp: 1548768231486,
};

type ObjectValueType = ValueOf<typeof obj>;
//   ^? string | number

const obj2 = ""
type Ex2 = ValueOf<typeof obj2> // list out all props of string domain set
// ^? string | number | (() => IterableIterator<string>) | (() => string) | ...

```

Cases like this are not handled in the implementation and it might lead to some strange results and affect the `ValueOf` type being used as well...

So what should be the Output for such Inputs which are in the Domain of Primitives?

One way can be to make it act like an Identity function. 
That is if the `ValueOf` is being passed a Primitive we can simply return that type.
Hence the Mapping will be `(D -> R)` and would cover all bases.

Proposed Solution - 

```ts
type Primitive = string | number | boolean | bigint | symbol | undefined | null;

type ValueOf<T> = T extends Primitive ? T: T[keyof T]

let obj1 = {name: "bishwajit", age: 22}

type Ex1 = ValueOf<typeof obj1>
    // ^? string | number

let obj2 = "hey"

type Ex2 = ValueOf<typeof obj2>
  // ^? string

const obj3 = "hey"

type Ex3 = ValueOf<typeof obj3>
   // ^? "hey"
 

```
 
 [Code Playground](https://www.typescriptlang.org/play?#code/C4TwDgpgBACgTgSwLYOAgbtAvFAzsRAOwHMoAfKQgVyQCMI5ypaB7FgGwgENCnaFiCQsCa4QdDkyqEAJhABmQiDKbV27ANwBYAFC7QkKADUu7KhADy8gDwAVAHxQctqBAAewCLNyxEKNJhQAPxQtgBcoQDaANYQICzyoQC6urqcIiy0AFYAjE5QAN6EXEgQEQBE-LgAFgDuXFmo5QA0UFzEZVAATF0Avqk6BtAAom55OCZmljZDCVCZufa6UCtQAPRrUAB6QQPp89ld+eXVceUDQ1CjRxOm5lbWs4kLXUs6Kxvbu3o6AMYshHwByyAGZjqcQOcfpdRmDblMHk9gSC3qtPjtlkA)
 
 
 Let me know what are your thoughts @Beraliv .
 If looks good I can raise a PR

ajitjha393:
Added the basic test coverage for `ValueOf<T>`.
@Beraliv Feel free to add more test cases or if we need to check the behavior for other edge cases as well. 

fixes #333

### DIFF
--- a/lib/types.ts
+++ b/lib/types.ts
@@ -10,6 +10,7 @@ export type IsAny<T> = 0 extends 1 & T ? true : false;
 export type IsNever<T> = [T] extends [never] ? true : false;
 export type IsUnknown<T> = IsAny<T> extends true ? false : unknown extends T ? true : false;
 export type AnyArray<T = any> = Array<T> | ReadonlyArray<T>;
+export type AnyFunction<TArgs extends any[] = any[], TReturnType = any> = (...args: TArgs) => TReturnType;
 
 export type ArrayOrSingle<T> = T | T[];
 
@@ -426,9 +427,9 @@ export type Opaque<Type, Token extends string> = Token extends StringLiteral<Tok
 /** Easily extract the type of a given object's values */
 export type ValueOf<T> = T extends Primitive
   ? T
-  : T extends any[]
+  : T extends AnyArray
   ? T[number]
-  : T extends (...args: any[]) => any
+  : T extends AnyFunction
   ? ReturnType<T>
   : T[keyof T];
 
--- a/lib/types.ts
+++ b/lib/types.ts
@@ -424,7 +424,13 @@ export type Opaque<Type, Token extends string> = Token extends StringLiteral<Tok
   : never;
 
 /** Easily extract the type of a given object's values */
-export type ValueOf<T> = T[keyof T];
+export type ValueOf<T> = T extends Primitive
+  ? T
+  : T extends any[]
+  ? T[number]
+  : T extends (...args: any[]) => any
+  ? ReturnType<T>
+  : T[keyof T];
 
 /** Easily extract the type of a given array's elements */
 export type ElementOf<T extends readonly any[]> = T extends readonly (infer ET)[] ? ET : never;
--- a/test/value-of.ts
+++ b/test/value-of.ts
@@ -1,0 +1,34 @@
+﻿import { AssertTrue as Assert, IsExact } from "conditional-type-checks";
+import { ValueOf, Primitive } from "../lib/types";
+
+declare const array: number[] | undefined;
+declare const func: (...arg: any[]) => boolean;
+declare const obj: { propA: string; propB: number; propC: () => void };
+declare const tuple: [1, 2, 3, 4];
+declare const primitive: Primitive;
+declare const primitiveOrObject: Primitive | typeof obj;
+declare const objOrFunc: typeof func | typeof obj;
+
+function testValueOf() {
+  type cases = [
+    Assert<IsExact<ValueOf<typeof obj>, string | number | (() => void)>>,
+    Assert<IsExact<ValueOf<typeof array>, number | undefined>>,
+    Assert<IsExact<ValueOf<typeof func>, boolean>>,
+    Assert<IsExact<ValueOf<typeof tuple>, 1 | 2 | 3 | 4>>,
+    Assert<IsExact<ValueOf<string>, string>>,
+    // valueOf for constant primitive literals will be the literal type
+    Assert<IsExact<ValueOf<"const">, "const">>,
+    Assert<IsExact<ValueOf<23>, 23>>,
+    Assert<IsExact<ValueOf<undefined>, undefined>>,
+    Assert<IsExact<ValueOf<{ a: "1"; b: 2 }>, "1" | 2>>,
+    // Combination of these types
+    Assert<IsExact<ValueOf<typeof primitive>, Primitive>>,
+    Assert<IsExact<ValueOf<typeof primitiveOrObject>, ValueOf<Primitive> | ValueOf<typeof obj>>>,
+    Assert<IsExact<ValueOf<typeof objOrFunc>, ValueOf<typeof obj | ValueOf<typeof func>>>>,
+    Assert<IsExact<ValueOf<23 | number>, number>>,
+    Assert<IsExact<ValueOf<[number[], string[]]>, number[] | string[]>>,
+    Assert<IsExact<ValueOf<number[] | string | null>, number | string | null>>,
+    Assert<IsExact<ValueOf<boolean[] | typeof obj>, boolean | ValueOf<typeof obj>>>,
+    Assert<IsExact<ValueOf<[string, number] | { keyA: bigint; keyB: null }>, string | number | bigint | null>>,
+  ];
+}
--- a/test/value-of.ts
+++ b/test/value-of.ts
@@ -3,11 +3,13 @@ import { ValueOf, Primitive, AnyArray } from "../lib/types";
 
 declare const array: number[];
 declare const func: (...arg: any[]) => boolean;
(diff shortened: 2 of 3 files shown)


Beraliv:
Hey again @ajitjha393!

Thank you for the help and your PR!

I will have a look closer next week, please bear with me!

lib/types.ts
```diff
@@ -424,7 +424,13 @@ export type Opaque<Type, Token extends string> = Token extends StringLiteral<Tok
   : never;
 
 /** Easily extract the type of a given object's values */
-export type ValueOf<T> = T[keyof T];
+export type ValueOf<T> = T extends Primitive
+  ? T
+  : T extends any[]
```
Beraliv, line 429: You check only writable array and tuples, but forgot about readonly arrays:

1. Please add the test for readonly array, e.g. `Assert<IsExact<ValueOf<readonly number[]>, number | undefined>>,`

2. Find a way to distinguish arrays, readonly arrays and tuples

3. Add the check here so tests are passed successfully

Let me know if you need any help for building conditional types

test/value-of.ts
```diff
@@ -0,0 +1,34 @@
...
+    Assert<IsExact<ValueOf<23>, 23>>,
+    Assert<IsExact<ValueOf<undefined>, undefined>>,
+    Assert<IsExact<ValueOf<{ a: "1"; b: 2 }>, "1" | 2>>,
+    // Combination of these types
+    Assert<IsExact<ValueOf<typeof primitive>, Primitive>>,
+    Assert<IsExact<ValueOf<typeof primitiveOrObject>, ValueOf<Primitive> | ValueOf<typeof obj>>>,
+    Assert<IsExact<ValueOf<typeof objOrFunc>, ValueOf<typeof obj | ValueOf<typeof func>>>>,
+    Assert<IsExact<ValueOf<23 | number>, number>>,
+    Assert<IsExact<ValueOf<[number[], string[]]>, number[] | string[]>>,
+    Assert<IsExact<ValueOf<number[] | string | null>, number | string | null>>,
```
Beraliv, line 30: You missed `| undefined` here as the `ValueOf<number[]>` is `number | undefined`

```suggestion
    Assert<IsExact<ValueOf<number[] | string | null>, number | undefined | string | null>>,
```

test/value-of.ts
```diff
@@ -0,0 +1,34 @@
...
+    Assert<IsExact<ValueOf<boolean[] | typeof obj>, boolean | ValueOf<typeof obj>>>,
```
Beraliv, line 31: Same here, you missed `| undefined` because of `ValueOf<boolean[]>`

lib/types.ts
```diff
@@ -424,7 +424,13 @@ export type Opaque<Type, Token extends string> = Token extends StringLiteral<Tok
...
+  ? T[number]
+  : T extends (...args: any[]) => any
```
Beraliv, line 431: If possible, could you please move it to `AnyFunction`, e.g.

```ts
type AnyFunction<TArgs extends any[] = any[], TReturnType = any> = (...args: TArgs) => TReturnType;
```

and it becomes:

```suggestion
  : T extends AnyFunction
```

lib/types.ts
```diff
@@ -424,7 +424,13 @@ export type Opaque<Type, Token extends string> = Token extends StringLiteral<Tok
   : never;
 
 /** Easily extract the type of a given object's values */
-export type ValueOf<T> = T[keyof T];
+export type ValueOf<T> = T extends Primitive
```
Beraliv, line 427: Please add changelog with `yarn changeset`

I imagine it will be patch as it's fixing existing utility function

ajitjha393:
Sure @beraliv, I'll iterate over the comments and suggestions and will do the required changes when I get some time after work
Thanks!

lib/types.ts
ajitjha393: > You check only writable array and tuples, but forgot about readonly arrays:

Oh yes totally missed that part, will change the conditional type to `AnyArray` as it will handle both mutable and readonly arrays.

> Find a way to distinguish arrays, readonly arrays, and tuples

I think in the conditional type `AnyArray` covers the case for `Tuple` as well because `ValueOf` should yield `T[number]` for all of them.
Have also added the test cases to distinguish b/w Arrays, readonly arrays, and tuples

```ts
export type ValueOf<T> = T extends Primitive
  ? T
  : T extends AnyArray
  ? T[number]
  : T extends AnyFunction
  ? ReturnType<T>
  : T[keyof T];
````

File: lib/types.ts (hunk 5 above)
Comment (ajitjha393) [reply lib/types.ts:427]: Makes sense, added the changeset summary and will update the PR

File: lib/types.ts (hunk 4 above)
Comment (ajitjha393) [reply lib/types.ts:431]: Yes will add this type helper `AnyFunction` and refactor it

File: test/value-of.ts (hunk 2 above)
Comment (Beraliv) [reply test/value-of.ts:43]: You're right, sorry for confusion 😅

Suggested names look better, thank you for clarifying it, I'm happy to resolve this comment then!

File: lib/types.ts (hunk 1 above)
Comment (Beraliv) [reply lib/types.ts]: Thank you for adjusting the logic, LGTM!

And yes, we don't need `IsTuple` as the result for arrays, readonly arrays and tuples will be identical!

**Comment by Beraliv** [comment]
@ajitjha393 thank you for the help, Bishwajit! I will need to extra-check how it works with `noUncheckedIndexedAccess` and  `exactOptionalPropertyTypes`, but all in all, great work!

File: lib/types.ts (hunk 1 above)
Comment (ajitjha393) [reply lib/types.ts]: Awesome!

**Comment by ajitjha393** [comment]
Cheers @Beraliv!
Glad I was able to help. 

> I will need to extra-check how it works with noUncheckedIndexedAccess and exactOptionalPropertyTypes

Sure, let me know when you are done with checking and we can merge it 👍

File: test/value-of.ts [hunk 6]
```diff
@@ -0,0 +1,47 @@
+﻿import { AssertTrue as Assert, IsExact } from "conditional-type-checks";
+import { ValueOf, Primitive, AnyArray } from "../lib/types";
+
+declare const array: number[];
+declare const func: (...arg: any[]) => boolean;
+declare const obj: { propA: string; propB: number; propC: () => void };
```
Beraliv, line 6: Can you please add 2 other tests with readonly and optional properties?

* existing obj will be `writableObj`
* `optionalObj` with optional properties
* `readonlyObj` with readonly properties

e.g. like in Playground – https://tsplay.dev/mAre1W

Beraliv (approved):
All good from my side! Thank you again for your help!

test/value-of.ts
ajitjha393: Sure!

ajitjha393:
Awesome!
I added 2 more test cases for the `readonly` and `optional` field object.
I'm ready to merge it. 
But I don't have write access, so you would have to merge it on my behalf. LGTM :D

Thanks, @Beraliv for helping out throughout the process!

Beraliv:
@ajitjha393 Thank you again for the help, I'm merging it now!

PR #336, opened 2022-11-08
fix the object constraint for `NonEmptyObject<T>`

### LINKED ISSUE #335: `NonEmptyObject<T>` type helper should have a constraint over an Object Type only for consistent result
Hey @Beraliv,
Sorry for bothering you repeatedly :D
 

### Current Implementation for `NonEmptyObject<T>`  

```ts
type NonEmptyObject<T extends {}> = keyof T extends never ? never : T
```

Here as we can see there is a problem in the constraint that we have applied on the type parameter `T extends {}` 
which allows even the Primitive types to be passed on and hence the consistency of the type is not preserved.

### Problem

The above type instantiation should give an error and not resolve to a Primitive type Output

```ts

type T0 = NonEmptyObject<number> 
//   ^? number

type T1 = NonEmptyObject<string> 
//    ^? string


```

Code Playground - [https://tsplay.dev/WPpAkW](https://tsplay.dev/WPpAkW)

ajitjha393:
fixed #335 

@Beraliv 
Have also added the additional test cases for the improved constraint!

### DIFF
--- a/.changeset/great-hounds-brush.md
+++ b/.changeset/great-hounds-brush.md
@@ -2,4 +2,4 @@
 "ts-essentials": patch
 ---
 
-improve the object constraint for `NonEmptyObject<T>` to not allow primitives
+Improve the object constraint for `NonEmptyObject<T>` to not allow primitives
--- a/.changeset/great-hounds-brush.md
+++ b/.changeset/great-hounds-brush.md
@@ -1,0 +1,5 @@
+---
+"ts-essentials": patch
+---
+
+improve the object constraint for `NonEmptyObject<T>` to not allow primitives
--- a/lib/types.ts
+++ b/lib/types.ts
@@ -385,7 +385,7 @@ type DeepModify<T> =
 /** Remove keys with `never` value from object type */
 export type NonNever<T extends {}> = Pick<T, { [K in keyof T]: T[K] extends never ? never : K }[keyof T]>;
 
-export type NonEmptyObject<T extends {}> = keyof T extends never ? never : T;
+export type NonEmptyObject<T extends Record<PropertyKey, any>> = keyof T extends never ? never : T;
 
 export type NonEmptyArray<T> = [T, ...T[]];
 
--- a/lib/types.ts
+++ b/lib/types.ts
@@ -385,7 +385,7 @@ type DeepModify<T> =
 /** Remove keys with `never` value from object type */
 export type NonNever<T extends {}> = Pick<T, { [K in keyof T]: T[K] extends never ? never : K }[keyof T]>;
 
-export type NonEmptyObject<T extends Record<PropertyKey, any>> = keyof T extends never ? never : T;
+export type NonEmptyObject<T extends AnyRecord> = keyof T extends never ? never : T;
 
 export type NonEmptyArray<T> = [T, ...T[]];
 
--- a/test/index.ts
+++ b/test/index.ts
@@ -15,7 +15,6 @@ import {
   DictionaryValues,
   Merge,
   MergeN,
-  NonEmptyObject,
   NonNever,
   noop,
   PickProperties,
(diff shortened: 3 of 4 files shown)


test/index.ts
```diff
@@ -836,8 +837,20 @@ function testNonEmptyObject() {
   type ObjectWithKeys = { foo: string; bar: number; xyz: undefined };
   type EmptyObject = {};
 
-  type TestA = Assert<IsExact<NonEmptyObject<ObjectWithKeys>, ObjectWithKeys>>;
-  type TestB = Assert<IsExact<NonEmptyObject<EmptyObject>, never>>;
+  type Cases = [
```
Beraliv, line 840: If you have time, as you've added more tests, could you please move them to a separate file, e.g. `non-empty-object.ts`?

If not, feel free to resolve this comment!

lib/types.ts
```diff
@@ -385,7 +385,7 @@ type DeepModify<T> =
 /** Remove keys with `never` value from object type */
 export type NonNever<T extends {}> = Pick<T, { [K in keyof T]: T[K] extends never ? never : K }[keyof T]>;
 
-export type NonEmptyObject<T extends {}> = keyof T extends never ? never : T;
+export type NonEmptyObject<T extends Record<PropertyKey, any>> = keyof T extends never ? never : T;
```
Beraliv, line 388: `PropertyKey` is not friendly for `keyofStringsOnly` option in tsconfig.json so we implemented our own

There's also a link attached to it, please have a look at it for more information – https://devblogs.microsoft.com/typescript/announcing-typescript-2-9-2/#support-for-symbols-and-numeric-literals-in-keyof-and-mapped-object-types

```suggestion
export type NonEmptyObject<T extends Record<KeyofBase, any>> = keyof T extends never ? never : T;
```

Beraliv, line 388: `PropertyKey` is not friendly for `keyofStringsOnly` option in tsconfig.json so we implemented our own

There's also a link attached to it, please have a look at it for more information – https://devblogs.microsoft.com/typescript/announcing-typescript-2-9-2/#support-for-symbols-and-numeric-literals-in-keyof-and-mapped-object-types

Also you can use `AnyRecord` for it as it's already defined in the file:

```suggestion
export type NonEmptyObject<T extends AnyRecord> = keyof T extends never ? never : T;
```

Beraliv (changes requested):
@ajitjha393 Thank you for your time again, Bishwajit!

I left one important comment about `PropertyKey`, please change it and I'm happy to approve it

test/index.ts
ajitjha393: Sure will refactor and move it to separate test file!

lib/types.ts
ajitjha393: I didn't know about this.
TIL :D
Went through the article, and it makes sense to have a separate `KeyofBase` type to not run into problems due to `keyofStringsOnly` compiler option.

And since we already have this `AnyRecord` implementation better to use it.

ajitjha393:
Pushed the required changes.
 `Record<PropertyKey, any>` to `AnyRecord` 
 Also refactored the test case for `NonEmptyObject` to a separate test file

Beraliv:
> Pushed the required changes.
>  `Record<PropertyKey, any>` to `AnyRecord` 
>  Also refactored the test case for `NonEmptyObject` to a separate test file

I'm back today, will have a look at it later today!

Beraliv (approved):
LGTM! Thank you for refactoring and adding more tests!

.changeset/great-hounds-brush.md
```diff
@@ -0,0 +1,5 @@
+---
+"ts-essentials": patch
+---
+
+improve the object constraint for `NonEmptyObject<T>` to not allow primitives
```
Beraliv, line 5:
```suggestion
Improve the object constraint for `NonEmptyObject<T>` to not allow primitives
```

Beraliv:
Released in v9.3.1 ✅

PR #346, opened 2023-01-27
Tweak `Tuple` type

Andarist:
This is probably a little bit saner way of declaring this constraint. The union trick is neat but a little bit hacky - the revised version uses the actual tuple type (without mixing it with an array) so it idiomatically makes sense why it works (well, I also understand why the union works as well but it always "looked off" to me). 

Arrays stay assignable to this and all tuples should still stay assignable to it (except the ones containing `never` types but that wasn't quite supported before either). 

Conceptually this should also work `[...any[]]` but unfortunately, this "simplifies" to just `any[]`, and by the time TS checks if the contextual type might be a tuple - it no longer is.

Feel free to close this PR, I mostly just opened it to have a reference to this thing - so I will always know where to find this 😅

### DIFF
--- a/lib/types.ts
+++ b/lib/types.ts
@@ -437,7 +437,7 @@ export type ValueOf<T> = T extends Primitive
 export type ElementOf<T extends readonly any[]> = T extends readonly (infer ET)[] ? ET : never;
 
 /** Type constraint for tuple inference */
-export type Tuple<T = any> = [T] | T[];
+export type Tuple<T = any> = [T?, ...T[]];
 
 /** Useful as a return type in interfaces or abstract classes with missing implementation */
 export type AsyncOrSync<T> = PromiseLike<T> | T;


Beraliv:
Hey @Andarist!

Thank you for the contribution, it makes total sense for me!

I didn't have a chance of updating `CONTRIBUTING.md` and adding template for PRs and issues, could you please add changeset using `yarn changeset`?

Otherwise it looks good and I'm happy to merge it!

Andarist:
Sure, i will happily add a changeset when i get home. You might want to install a changeset bot to post helpful reminders in PRs :)

Andarist:
Added the requested changeset

Beraliv (approved):
Thank you for the change, @Andarist! I will include it in next release – https://github.com/ts-essentials/ts-essentials/pull/337

Beraliv:
Released in v9.3.1 ✅

PR #348, opened 2023-02-11
fix: Opaque that works

ukstv:
Apparently, current upstream implementation of `Opaque` is unusable. It leads to this error: https://github.com/microsoft/TypeScript/issues/37888

This PR changes slightly the code, so that now consumers of `ts-essential` can actually use `Opaque` type.

### DIFF
--- a/lib/types.ts
+++ b/lib/types.ts
@@ -419,10 +419,12 @@ type StringLiteral<T> = T extends string ? (string extends T ? never : T) : neve
 
 declare const __OPAQUE_TYPE__: unique symbol;
 
+export type WithOpaque<Token extends string> = {
+  readonly [__OPAQUE_TYPE__]: Token;
+};
+
 /** Easily create opaque types ie. types that are subset of their original types (ex: positive numbers, uppercased string) */
-export type Opaque<Type, Token extends string> = Token extends StringLiteral<Token>
-  ? Type & { readonly [__OPAQUE_TYPE__]: Token }
-  : never;
+export type Opaque<Type, Token extends string> = Token extends StringLiteral<Token> ? Type & WithOpaque<Token> : never;
 
 /** Easily extract the type of a given object's values */
 export type ValueOf<T> = T extends Primitive


Beraliv:
Hey @ukstv 👋

Thank you for your contribution! I really appreciate your help 🧡

As far as I understand you have `Exported type 'BlaBla' has or is using name '__OPAQUE_TYPE__' from external module "ts-essentials/lib/types.ts" but cannot be named. ts(4023)` error at the moment, right?

Could you please:

- Add the test for this scenario to be able to handle it next time (there are enough examples to reproduce it, any of them will are good enough)
- Add changelog using `yarn changeset` to explain the issue that you're fixing, e.g. `` Fix unique symbol `__OPAQUE_TYPE__` in `Opaque` that cannot be used outside of ts-essentials ``
- Update PR title because we "Squash and merge" changes, it will be unclear what "fix: Opaque that works" means, I suggest to use something like "Fix unique symbol in Opaque that cannot be used outside of ts-essentials"

Let me know if you need any help!

Beraliv:
Hey @ukstv! Kind reminder about this PR ⏰

Are you happy to do it yourself?

Otherwise I will close this PR as stale, create an issue and will resolve it myself.

Let me know what you think

ukstv:
Do not close this yet, please.

ukstv:
- [x] Proper PR title
- [x] Add test
- [x] `yarn changeset`

There is one issue though @Beraliv `--noEmit` does not catch the issue. TSC only complains about the "symbol that can not be named" if it actually builds the code.

To see it by yourself, copy [opaque.ts test](https://github.com/ts-essentials/ts-essentials/blob/f34bb05aca5cab6f8f0f10d363d4d9583d10008b/test/opaque.ts) into your upstream `test` folder, and run  `yarn test`. It would succeed. Then, remove `--noEmit`, run again. You'll see TS complaining now.

I changed `test` script so that it emits the code to `/dist`. Please, let me know how that could be done better.

Beraliv:
> * [x]  Proper PR title
> * [x]  Add test
> * [x]  `yarn changeset`
> 
> There is one issue though @Beraliv `--noEmit` does not catch the issue. TSC only complains about the "symbol that can not be named" if it actually builds the code.
> 
> To see it by yourself, copy [opaque.ts test](https://github.com/ts-essentials/ts-essentials/blob/f34bb05aca5cab6f8f0f10d363d4d9583d10008b/test/opaque.ts) into your upstream `test` folder, and run `yarn test`. It would succeed. Then, remove `--noEmit`, run again. You'll see TS complaining now.
> 
> I changed `test` script so that it emits the code to `/dist`. Please, let me know how that could be done better.

Thank you for the fix! I approved the run to see CI checks

I will also revisit it later today, will try to test it myself

Beraliv:
Great job @ukstv! I'm happy with your changes!

The only 2 things I'd like to suggest before approving it:

1. **Add `outDir` parameter for `test` script**

```diff
- "test": "rimraf ./dist && prettier -c **/*.ts && tsc --noEmit",
+ "test": "rimraf ./testDist && prettier -c **/*.ts && tsc --outDir ./testDist",
```

To have better DX: after your changes it will be hard to distinguish developer changes from emitted by `tsc`, currently you have to clean emitted files and commit only necessary changes which isn't convenient

Also you clean `dist` after each test, it should be `testDist` (otherwise you will clean built files)

2. **Add `outDir` value to `.gitingore`**

Emitted files don't need to be committed, so it's cleaner to hide them for git repo

I've tested it locally, it works quite well for me,

ukstv:
Okay. Doing it now.

ukstv:
@Beraliv Is it better now?

Beraliv (approved):
Thank you very much! I'm happy to merge it!

Beraliv:
Released in v9.3.1 ✅

PR #351, opened 2023-02-22
feat: Create mechanism to add generic Deep types

Beraliv:
**Summary**

- Added `Fn` and `Call` (based on hotscript)
- Added **Extractors** based on the conditional types we have (11 + else-branch = 12 in total)
- Added **Resolvers** to resolve the return type based on the deep type

TODO:

- [ ] Fix the issue with `Type instantiation is excessively deep and possibly infinite`
- [ ] `DeepReadonly`
- [ ] `DeepPartial`
- [ ] `DeepWritable`
- [ ] `DeepRequired`

**Why**

To minimise the repeats in all deep-types

Related to https://github.com/ts-essentials/ts-essentials/issues/12

cc the idea is adopted from hotscript (cc @gvergnaud @ecyrbe). I want to contribute to it in the end so any help would be appreciated


Beraliv:
Closing it for now as it's not prioritised

Issue #349, opened 2023-02-17
Assignability for MarkRequired after upgrade from 9.1.2 to 9.3.0


akwodkiewicz:
Similarly to #323, my type guard is not working correctly after the change inside `MarkRequired` (`Exclude` => `Omit`). What's more is that the addition of union handling (done in 9.3.0) has not fixed the problem. Here's the type guard with the error.

```ts
export function checkSetProperties<T, PropName extends keyof T>(
    obj: T,
    propNames: PropName[],
): obj is MarkRequired<T, PropName> {
//        ^^^^^^^^^^^^^^^^^^^^^^^^^
//        A type predicate's type must be assignable to its parameter's type.
//          Type 'MarkRequired<T, PropName>' is not assignable to type 'T'.
//            'T' could be instantiated with an arbitrary type which could be unrelated to 'MarkRequired<T, PropName>'.ts(2677)
    return !propNames.some(prop => !obj[prop]);
}
```

[...]

PR #364, opened 2023-04-02
fix: Assignability for Mark* utility types

Beraliv:
## PR Checklist

- [x] Addresses an existing open issue: related to #349
- [ ] Steps in [Contributing](https://github.com/ts-essentials/ts-essentials/blob/master/CONTRIBUTING.md) were taken

## Overview

Assignability of Mark* utility types, e.g. `MarkRequired` should be identical to its builtin type `Required`

- Added `Debug` type for tests
- Added assignability checks – https://github.com/ts-essentials/ts-essentials/commit/ef16a24f37a822c1f51b03b2253322dce39dafe3
- Fixed 3 utility types `MarkRequired`, `MarkReadonly` and `MarkWritable`

### DIFF
--- a/lib/mark-readonly/index.ts
+++ b/lib/mark-readonly/index.ts
@@ -1,3 +1,7 @@
+import { ReadonlyKeys } from "../readonly-keys";
+import { Writable } from "../writable";
+
 export type MarkReadonly<Type, Keys extends keyof Type> = Type extends Type
-  ? Omit<Type, Keys> & Readonly<Pick<Type, Keys>>
+  ? Readonly<Type> &
+      Writable<Pick<Type, Exclude<keyof Type, Keys | (Type extends object ? ReadonlyKeys<Type> : never)>>>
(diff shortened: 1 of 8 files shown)


akwodkiewicz:
So it looks like the previous approach of "subtracting" the properties that we want to make required/writable/readonly was wrong and now we... overwrite the existing properties with type intersection? 😄

Beraliv:
> So it looks like the previous approach of "subtracting" the properties that we want to make required/writable/readonly was wrong and now we... overwrite the existing properties with type intersection? 😄

I'm trying to understand the reason and I see the difference between different TypeScript versions now. Looks like a lot of things were fixed in TypeScript 4.5, so I will probably bump the minimum version and will simplify the solution so it will be correctly working for all cases

But at least it's clear that previous approach wasn't working for both assignability and correct `readonly` properties because writable properties absorb `readonly` properties

PR #367, opened 2023-04-05
fix: support types extending Array in DeepRequired

### LINKED ISSUE #366: Bug: [DeepRequired] Type errors when using `Array` interface extension
### Versions

| package                                | version |
| -------------------------------------- | ------- |
| `typescript`                           | `5.0.3` |
| `ts-essentials`                        | `9.3.1` |

### Repro Code

See TS playground, which includes `DeepRequired` from ts-essentials 9.3.1. Note the potentially naive fix suggested in the copied ts-essentials code, as well as the workaround at the bottom.

### TypeScript Playground

[here](https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAbzgIQK7ADY2AOzgXzgDMoIQ4ByGAZwFoBTa6+nbAQw2ooG4AoXgPQC4AAwAi9emABK9AI7oo9ACYjipcgDcAnADoAzLoCMvegA9IsOMvoBjDGyVwYATzD04EqbIXAlygB4AFQA+OABeOCC4cxgWZWo4AFEoUiheODgAfjgfRRVgkIy4AC4omLM4nASUdCxcYpyg4qE4AAUIKvYMYmAzMtQcWzIQFngYAHcIOAxcRjgAI3oMCAnnaaVqCAxND3pU6BbhMujY+MSAQVS2FwDBgGscVZwizNam4pOKqpqAWTYwAFcER9nAANIAGjgwNBADVXtk4P9AV4ZPJ8oEwSEoai8n4CvCEV8ztVErI2MoIDgMC5kUCcCCoOCoTCmfDGrl6BSqTS6bj0fjMdjPJI0b5-AFCZ9yiSagB1Ln3Oms5nQhlwhE5BVsJUAgL88UFLE40V4iVSzLEyrnOAAZXoMHpjLgAFVNXaHfrTQKJW6iTLraTOdzqS57Y6VW6OeTKaHw17vD6Cn7padA-LFfHI+7tfd4waMQEU5aAz9Em0NMBmE7QVHMjkK2Qq-QE2LC8XSqWbQh8ByEMVMpkANpgtVwe70FwQIhRAC6tCyZQLguCI9nIT4g4I0rNBVCfF4uDiUCIbFsHlhHFQ9CuUBu3xtt5uAWoMCguAA5mEe-xhjhX3AABiEAQEu3qGoESALGwABei5wJeGDXk+LgEGEkT9lu0EwWUQ4UAoqBmBQs4DnArRytA9yOBAgzKCUpGtNhuH4aghHEXAbCJMu5pXje1wuNivD4NwQA)

### Expected Result

It should be possible to use `DeepRequired` with objects containing properties whose value is of a type which extends `Array<T>`. In situations like this, the property should be required to be set, but the expected members of the array should not change.

### Actual Result

In TS ~4, this works as expected. However, in TS5.0.3, there's a compilation error - it looks to be down to the fact that the DeepRequired type is modifying the members of the Array interface:

```
The types of '[Symbol.unscopables].length' are incompatible between these types.
    Type 'boolean | undefined' is not assignable to type 'boolean'.
      Type 'undefined' is not assignable to type 'boolean'.(2322)
```

mtlewis:
## PR Checklist

- [x] Addresses an existing open issue: related to #366

## Overview

Work-in-progress attempt to fix #366. See comments inline.

### DIFF
--- a/lib/deep-required/index.ts
+++ b/lib/deep-required/index.ts
@@ -1,4 +1,5 @@
 import { Builtin } from "../built-in";
+import { IsTuple } from "../is-tuple";
 
 export type DeepRequired<Type> = Type extends Error
   ? Required<Type>
@@ -18,6 +19,12 @@ export type DeepRequired<Type> = Type extends Error
   ? WeakSet<DeepRequired<Values>>
   : Type extends Promise<infer Value>
   ? Promise<DeepRequired<Value>>
+  : Type extends ReadonlyArray<infer Values>
+  ? Type extends IsTuple<Type>
+    ? { [Key in keyof Type]-?: DeepRequired<Type[Key]> }
+    : Type extends Array<Values>
+    ? Array<Exclude<DeepRequired<Values>, undefined>>
+    : ReadonlyArray<Exclude<DeepRequired<Values>, undefined>>
   : Type extends {}
   ? { [Key in keyof Type]-?: DeepRequired<Type[Key]> }
   : Required<Type>;
--- a/test/deep-required.ts
+++ b/test/deep-required.ts
@@ -19,6 +19,10 @@ function testDeepRequired() {
     caseSensitive: true;
   }
 
+  interface ExtendedArray<T> extends Array<T> {}
+
+  interface ExtendedReadonlyArray<T> extends ReadonlyArray<T> {}
+
   type cases = [
     Assert<IsExact<DeepRequired<number | null | undefined>, number | null | undefined>>,
     Assert<IsExact<DeepRequired<string | null | undefined>, string | null | undefined>>,
@@ -89,6 +93,9 @@ function testDeepRequired() {
     >,
     Assert<IsExact<DeepRequired<readonly (number | null | undefined)[]>, readonly (number | null)[]>>,
     Assert<IsExact<DeepRequired<Array<number | null | undefined>>, Array<number | null>>>,
+    Assert<IsExact<DeepRequired<ReadonlyArray<number | null | undefined>>, ReadonlyArray<number | null>>>,
+    Assert<IsExact<DeepRequired<ExtendedArray<number | null | undefined>>, Array<number | null>>>,
+    Assert<IsExact<DeepRequired<ExtendedReadonlyArray<number | null | undefined>>, ReadonlyArray<number | null>>>,
     Assert<IsExact<DeepRequired<Promise<number | null | undefined>>, Promise<number | null | undefined>>>,
     Assert<
       IsExact<
--- a/.changeset/silly-buses-camp.md
+++ b/.changeset/silly-buses-camp.md
@@ -1,0 +1,5 @@
+---
+"ts-essentials": patch
+---
+
+Add support for types which explicitly extend Array inside types passed to `DeepRequired`.


lib/deep-required/index.ts
```diff
@@ -18,6 +20,12 @@ export type DeepRequired<Type> = Type extends Error
   ? WeakSet<DeepRequired<Values>>
   : Type extends Promise<infer Value>
   ? Promise<DeepRequired<Value>>
+  : Type extends AnyArray<infer Values>
```
mtlewis, line 23: It would be really nice to avoid Tuples leaking through to the left-hand side of the conditional so that we can let them drop all the way down to the type (now) on line 30 and reduce the repetition a bit.

lib/deep-required/index.ts
```diff
@@ -18,6 +20,12 @@ export type DeepRequired<Type> = Type extends Error
...
+  ? Type extends Tuple
+    ? { [Key in keyof Type]-?: DeepRequired<Type[Key]> }
+    : Type extends Array<Values>
```
mtlewis, line 26: It seems a real shame that we have to explicitly deal with both Array and ReadonlyArray here, but I'm not quite sure how to handle both types more implicitly.

lib/deep-required/index.ts
Beraliv: I think it's my bad that `AnyArray` is used in these utility types, you can simply use `ReadonlyArray<infer Values>` or `readonly (infer Values)[]`

For tuples, there's a utility type [IsTuple](https://github.com/ts-essentials/ts-essentials/tree/master/lib/is-tuple), you can use it to filter it and process only arrays at this point

lib/deep-required/index.ts
Beraliv: It's possible to unite it with another type, but for simplicity it's left as is so it will be clear what happens for new people

Sometimes we have to do different things for readonly, normal arrays and tuples, that's why we still do it manually

If you have any suggestion how to generally handle it better, I'm more than open to address it in another issue

lib/deep-required/index.ts
```diff
@@ -18,6 +20,12 @@ export type DeepRequired<Type> = Type extends Error
   ? WeakSet<DeepRequired<Values>>
   : Type extends Promise<infer Value>
   ? Promise<DeepRequired<Value>>
+  : Type extends AnyArray<infer Values>
+  ? Type extends Tuple
```
Beraliv, line 24: As I suggested above, you can use `IsNever<IsTuple<Type>>` so you will filter out tuple at this point so you don't have to duplicate the same code for objects and tuples

lib/deep-required/index.ts
```diff
@@ -1,4 +1,6 @@
+import { AnyArray } from "../any-array";
 import { Builtin } from "../built-in";
+import { Tuple } from "../tuple";
 
 export type DeepRequired<Type> = Type extends Error
```
Beraliv, line 5: You still lack the tests for the new changes, please have a look at other Deep-utility types and add tests for:

- readonly arrays
- normal arrays
- tuples

If there're already in place, you can add only missing ones

Beraliv (changes requested):
Thank you for your contribution!

Good start, but still missing tests for tuples and types that extend Arrays

Let me know if you need any help

lib/deep-required/index.ts
mtlewis: Yep - mentioned this in the issue. Will be sure to add more tests.

mtlewis: Some of these were already in place, but I believe I've covered the missing ones. PTAL

lib/deep-required/index.ts
mtlewis: Done :+1:

lib/deep-required/index.ts
mtlewis: I definitely don't - was kinda hoping I was missing something 😁.

lib/deep-required/index.ts
mtlewis: Your suggestion isn't totally clear to me, so I opted instead to follow the pattern in deep-partial and duplicate the object code.

test/deep-required.ts
```diff
@@ -89,6 +93,9 @@ function testDeepRequired() {
     >,
     Assert<IsExact<DeepRequired<readonly (number | null | undefined)[]>, readonly (number | null)[]>>,
     Assert<IsExact<DeepRequired<Array<number | null | undefined>>, Array<number | null>>>,
+    Assert<IsExact<DeepRequired<ReadonlyArray<number | null | undefined>>, ReadonlyArray<number | null>>>,
+    Assert<IsExact<DeepRequired<ExtendedArray<number | null | undefined>>, Array<number | null>>>,
```
mtlewis, line 97: This case is slightly odd, in that types that extend Array get mapped back to `Array` as a side-effect of applying `DeepRequired` (and the same goes for types that extend `ReadonlyArray`).

This matches the existing behavior in deep-readonly, but it's less strange there, since the type is implicitly responsible for converting array types to `ReadonlyArray` types.

mtlewis:
@Beraliv thanks for the feedback - this is ready for another look.

Beraliv:
@mtlewis apologies for a very long review, let me have a look at this PR again this week and I will come back to you

Beraliv:
@mtlewis I'm once again sorry for a very long review, last months were quite busy. I will let you know when I have a look at this. Keep pinging me to get it merged

test/deep-required.ts
Beraliv: this is a side-effect of conditional types that you have added: 

```ts
Type extends Array<Values>
    ? Array<Exclude<DeepRequired<Values>, undefined>>
    : ReadonlyArray<Exclude<DeepRequired<Values>, undefined>>
```

Now it's mapped back to `Array` or `ReadonlyArray` depending on what it `extends`

test/types.ts
```diff
@@ -1,9 +1,13 @@
+interface ExtendedArray<T> extends Array<T> {}
+
 export type ComplexNestedRequired = {
   simple: number;
   nested: {
     date: Date;
     func: () => string;
     array: { bar: number }[];
+    readonlyArray: ReadonlyArray<{ bar: number }>;
+    extendedArray: ExtendedArray<{ bar: number }>;
```
Beraliv, line 10: the way those types are created in this file won't allow you to add an extended arrays so I suggest removing it

The reason is that all the complex nested types here are the results of calling `DeepRequired` therefore extended array type will be converted to Array type

test/types.ts
```diff
@@ -1,9 +1,13 @@
+interface ExtendedArray<T> extends Array<T> {}
```
Beraliv, line 1: I suggest reverting changes in this file because the following rule has to be true: `ComplexNestedRequired = DeepRequired<ComplexNestedPartial>` and `ComplexNestedRequired = DeepWritable<ComplexNestedReadonly>`

readonly and extended arrays are an exception to these rules so has to be handled separately

Beraliv:
@mtlewis Thank you for your patience! I've just reviewed your changes. All main changes look good, thank you for adding `Exclude` which keeps backward compatibility. Once tests are updated according to my comments, I'm happy to approve and merge the PR.

test/types.ts
mtlewis: Done!

mtlewis:
@Beraliv addressed your comment by reverting the changes to test/types.ts as suggested. This should be good to go from my perspective.

Beraliv (approved):
LGTM! Thank you for your patience and great work on resolving the issue 👏

Issue #380, opened 2024-01-20
Bug: [DeepPick] on nullable type


AndreasHald:
### Versions

| package                                | version |
| -------------------------------------- | ------- |
| `typescript`                           | `X.Y.Z` |
| `ts-essentials`                        | `X.Y.Z` |

### Repro Code

```typescript
import { DeepPick } from 'ts-essentials';

interface EnglishClass {
  teacher: {
    name: string;
    yearsOfExperience: number;
  } | null;
}

type EnglishClassNamesOnly = DeepPick<
  // ^?
  EnglishClass,
  {
    teacher: {
      name: never;
    };
  }
>;
```

### TypeScript Playground

https://www.typescriptlang.org/play?ssl=7&ssc=12&pln=8&pc=2#code/JYWwDg9gTgLgBAbzgEQKarABWAYwNZwC+cAZlBCHAOQwDOAtKrbagHYzACGANrVQNwBYAFAjg7VFBKccqOAFFWAc27BaACwDC3Ts0Qi4cGKhnrJALn3DDh1pxCpLtGFHFKh1mwE8TUWgHkSeQAPMElgNllLVgBXEAAjSQ9DYgAfOFjubg9CEREYLzCFZVUNbV1aADl7Jn9Wbi84AF4UdCxcPAAeAzgAel64AD0Afh7FFTUtHWYAGh6EHsNjUwsrGxs7B2jUADckxaJkohEAPg8gA

### Expected Result

I expected a type to be constructed that would be 
```ts 
teacher: {
name: string
} | null
```

### Actual Result

TS error

AndreasHald:
It actually looks like it constructs the type properly - however a TS error is still thrown

![CleanShot 2024-01-20 at 14 05 37@2x](https://github.com/ts-essentials/ts-essentials/assets/8914722/a39341a8-15fe-45d3-a037-285a50f0cc41)

Beraliv:
Hey @AndreasHald

Thank you for reporting the issue, it looks like `DeepPick` doesn't support nullable properties because it shows the generic constraint is `{ teacher?: true | undefined; }` and doesn't contain an option with partial `Teacher` object

I apologise in advance but I won't be available next 2 weeks. When I'm back, I can have a look at this

Beraliv:
Hey again @AndreasHald, looking at this issue right now

[...]

Issue #343, opened 2022-12-28
Unable to use DeepOmit with generic type


xenoterracide:
```ts
export type ProducerOptions = DeepOmit<ProducerDocument, EntityOmit>;

export type ProducerDocument = DeepOmit<
  // eslint-disable-next-line @typescript-eslint/ban-types
  OmitProperties<ProducerAggregateRoot, Function>,
  AggregateOmit
>;
```

```
Type 'AggregateOmit' does not satisfy the constraint 'DeepModify<T>'.
  Type 'AggregateOmit' is not assignable to type 'T extends WeakMap<infer K extends object, infer E> ? WeakMap<K, DeepModify<E>> : never'.ts(2344)
```

I'm repeatedly doing this

```ts
export interface Entity<Id> {
  id: Id;
  createdAt: Date;
  modifiedAt: Date;
}

export type EntityOmit = DeepReplace<Entity<never>, never>;
export type AggregateOmit = {
  autoCommit: never;
  [index: symbol]: never;
};
```

all extend [`AggregateRoot`](https://github.com/nestjs/cqrs/blob/master/src/aggregate-root.ts)

this doesn't work though

```ts
export type OmitIt<T extends AggregateRoot> = DeepOmit<
  // eslint-disable-next-line @typescript-eslint/ban-types
  OmitProperties<T, Function>,
  AggregateOmit
>;
```

Beraliv:
Hey!

Let me have a look at it quickly

Nightbr:
Same here, cannot be able to DeepOmit on generic:

```
import { DeepOmit } from 'ts-essentials';

export type ExposedData<T> = DeepOmit<
  T,
  {
    id: never;
    isLatest: never;
    checkId: never;
    createdAt: never;
    updatedAt: never;
  }
>;
```

```
Type '{ id: never; isLatest: never; checkId: never; createdAt: never; updatedAt: never; }' does not satisfy the constraint 'DeepModify<T>'.
  Type '{ id: never; isLatest: never; checkId: never; createdAt: never; updatedAt: never; }' is not assignable to type 'T extends WeakMap<infer K extends object, infer E> ? WeakMap<K, DeepModify<E>> : never'.ts(2344)
```

[...]

PR #384, opened 2024-04-06
feat: ➕ add StrictDeepPick and StrictDeepOmit

Beraliv:
## PR Checklist

- [x] Addresses an existing open issue: related to #343 and partially #380

## Overview

- added `StrictDeepPick` and `StrictDeepOmit` that support generic type
- ⚠️ change the implementation of `DeepPick` and `DeepOmit` which now don't have a generic constraint on second type parameter (i.e. it's not validated against the type anymore)
- removed `DeepModify` from `DeepPick` and `DeepOmit`

### DIFF
--- a/.changeset/unlucky-camels-smile.md
+++ b/.changeset/unlucky-camels-smile.md
@@ -1,5 +1,6 @@
 ---
-"ts-essentials": minor
+"ts-essentials": major
 ---
 
-Add `StrictDeepOmit` and `StrictDeepPick` that support generic type
+Added `StrictDeepOmit` and `StrictDeepPick` that support generic type and removed generic constraint on the second type
+parameter of `DeepOmit` and `DeepPick`
--- a/.changeset/unlucky-camels-smile.md
+++ b/.changeset/unlucky-camels-smile.md
@@ -1,0 +1,5 @@
+---
+"ts-essentials": minor
+---
+
+Add `StrictDeepOmit` and `StrictDeepPick` that support generic type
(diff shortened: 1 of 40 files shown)


.changeset/unlucky-camels-smile.md
```diff
@@ -0,0 +1,5 @@
+---
+"ts-essentials": minor
```
Beraliv, line 2: Given the convention that I've just added, it doesn't break the backwards compatibility so I'm happy to continue with minor

Although, if it's treating as a break of contract, it's changing the current behaviour (type error => no type error) therefore I'm also happy to make it major

Any input from anyone?

lib/deep-omit/index.ts
```diff
@@ -1,67 +1,46 @@
 import { AnyRecord } from "../any-record";
 import { Builtin } from "../built-in";
-import { DeepModify } from "../deep-modify";
 
-export type DeepOmit<Type, Filter extends DeepModify<Type>> = Type extends Builtin
+export type DeepOmit<Type, Filter> = Type extends Builtin
   ? Type
   : Type extends Map<infer Keys, infer Values>
   ? Filter extends Map<Keys, infer FilterValues>
-    ? FilterValues extends DeepModify<Values>
```
Beraliv, line 9: 9 conditional types were removed in `DeepOmit` and `DeepPick` which improves performance of these utility types after this change

.changeset/unlucky-camels-smile.md
Beraliv: After @Nightbr's comment, I've considered using major instead of minor - https://github.com/ts-essentials/ts-essentials/issues/343#issuecomment-2041623070

PR #394, opened 2024-05-18
Add .gitattibute for contribution from Windows

DanielBertocci:
## PR Checklist

- [x] Addresses an existing open issue: related to #000

## Overview
I have had issues for contributing: I am in a Windows environment and git converts line endings with CRLF (by default core.autocrlf is set to true); some script like test do not work if this conversion happens.

I have added .gitattribute to avoid this from happening.


Beraliv:
Hey @DanielBertocci!

Thanks a lot for your PR, very appreciate you've looked at the Windows configuration

I'm looking at your PR now. For future engineers benefit, I will add some comments for what each config setting means

.gitattribute
```diff
@@ -0,0 +1,6 @@
+# Auto detect text files and perform LF normalization
+* text=auto
```
Beraliv, line 2: > When text is set to "auto", the path is marked for automatic end-of-line normalization. If Git decides that the content is text, its line endings are normalized to LF on checkin.

Source - https://git-scm.com/docs/gitattributes/2.5.6#Documentation/gitattributes.txt-Settostringvalueauto

.gitattribute
```diff
@@ -0,0 +1,6 @@
...
+
+# JS and TS files must always use LF for tools to work
+*.js eol=lf
+*.ts eol=lf
```
Beraliv, line 6: @DanielBertocci I believe you need an empty line here:

```suggestion
*.ts eol=lf

```

Beraliv, line 6: > his setting forces Git to normalize line endings to LF on checkin and prevents conversion to CRLF when the file is checked out.

Source - https://git-scm.com/docs/gitattributes/2.5.6#Documentation/gitattributes.txt-Settostringvaluelf

Beraliv:
@DanielBertocci thanks a lot, merging it!

DanielBertocci:
@Beraliv I've noticed that I missed a s in the file name. It should be .gitattributes...it worked on my machine because in attempts to understand why I had those issues I changed the default behavior of git.

PR #395, opened 2024-05-18
Use key remapping from Typescript 4.1 for PickProperties and OmitProperties

DanielBertocci:
## PR Checklist

- [x] Addresses an existing open issue: related to #000

## Overview
I took advantage of [key remapping](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-1.html#key-remapping-in-mapped-types) feature introduced starting from Typescript 4.1.

This allows to reduce the number of iterations when these types are used and to remove the utility PickKeysByValue.

In tests, the "wtf" cases continue to be a bit "wtf" but they are closer to be keys, at least.

I wanted to use this syntax for other cases, but it changes the behavior a bit more and, despite I think it would be reasonable the new behavior, it requires more discussion.

### DIFF
--- a/lib/omit-properties/index.ts
+++ b/lib/omit-properties/index.ts
@@ -1,3 +1,1 @@
-import { PickKeysByValue } from "../pick-keys-by-value";
-
-export type OmitProperties<Type, Value> = Omit<Type, PickKeysByValue<Type, Value>>;
+export type OmitProperties<Type, Value> = { [Key in keyof Type as Type[Key] extends Value ? never : Key]: Type[Key] };
--- a/lib/pick-keys/index.ts
+++ b/lib/pick-keys/index.ts
@@ -1,3 +1,3 @@
 import { PickProperties } from "../pick-properties";
 
-export type PickKeys<Type, Value> = Exclude<keyof PickProperties<Type, Value>, undefined>;
+export type PickKeys<Type, Value> = keyof PickProperties<Type, Value>;
--- a/lib/pick-keys-by-value.ts
+++ b/lib/pick-keys-by-value.ts
@@ -1,1 +1,0 @@
-export type PickKeysByValue<Type, Value> = { [Key in keyof Type]: Type[Key] extends Value ? Key : never }[keyof Type];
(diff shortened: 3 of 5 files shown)


Beraliv:
@DanielBertocci Hey again! 👋

I really appreciate you took your time to improve some of the types. As the current `ts-essentials` version requires `typescript>=4.5`, I'm happy to apply these improvements.

But before merging the changes, let me quickly analyse what's the performance gain (the diff between the "before" and the "after" in terms of instantiations) because I think it's a significant improvement (FYI I'm doing benchmarks in this PR - https://github.com/ts-essentials/ts-essentials/pull/393)

lib/omit-properties/index.ts
```diff
@@ -1,3 +1 @@
-import { PickKeysByValue } from "../pick-keys-by-value";
-
-export type OmitProperties<Type, Value> = Omit<Type, PickKeysByValue<Type, Value>>;
+export type OmitProperties<Type, Value> = { [Key in keyof Type as Type[Key] extends Value ? never : Key]: Type[Key] };
```
Beraliv, line 1: Amazing work! It's very performant right now

Screenshot 2024-05-19 at 12 20 49

lib/pick-keys/index.ts
```diff
@@ -1,3 +1,3 @@
 import { PickProperties } from "../pick-properties";
 
-export type PickKeys<Type, Value> = Exclude<keyof PickProperties<Type, Value>, undefined>;
+export type PickKeys<Type, Value> = keyof PickProperties<Type, Value>;
```
Beraliv, line 3: Ha, it's amazing that key remapping fixed the problem with `undefined`

Beraliv (approved):
@DanielBertocci LGTM! I don't see any regressions as all tests are passed. I will update the instantiations and create changelog in a separate PR so I'm happy for this PR to be merged now

DanielBertocci:
Great! Just as remark that this PR introduces a breaking change on the edge cases:
- `PickKeys<boolean, number | undefined>` now maps to `"valueOf"`
- `PickKeys<symbol, number | undefined>` now maps to `SymbolConstructor["toPrimitive"] | SymbolConstructor["toStringTag"] | "toString" | "valueOf"`

Beraliv:
@DanielBertocci I am ready to accept the risk that this is a very unrealistic edge case therefore I will leave it as a "patch" anyway. The majority of the cases should apply this type to the object.

Beraliv:
@DanielBertocci This improvement has been under [v10.0.1](https://github.com/ts-essentials/ts-essentials/releases/tag/v10.0.1), thank you again for your help!

Issue #411, opened 2024-09-27
Enhancement: [Paths<Type>] Add configurable wildcard for matching array indices


mylesj:
### What

I propose an addition to [`Paths<Type>`](https://github.com/ts-essentials/ts-essentials/tree/master/lib/paths) - specifically the [options](https://github.com/ts-essentials/ts-essentials/blob/d0ad79f3e022f9ac97b5ec08314a20d3286db9dc/lib/paths/index.ts#L25-L42) - that would allow for a more permissive wildcard with regard to array indices. For context I'm writing a generic _"search"_ utility for an object store in my project, and would like for my consumer to be able to differentiate their interest in a specific element from interest in all elements of the same set.

### Examples

```typescript
interface Company {
  name: string;
  employees: { name: string }[];
}

const path: Paths<Company, { wildcard: '*' }> = 'employees.*.name';
```

### Additional Info

Strictly speaking I can already work around this by using something like `-1` as my wildcard, however this is obscure to the reader - I'd prefer that my API is explicit and discoverable. 

```typescript
const path: Paths<Company> = 'employees.-1.name'; // satisfies TypeScript but not idiomatic
```

Beraliv:
Hey @mylesj!

Thank you for raising an issue with your proposal!

I agree that adding an option to [options](https://github.com/ts-essentials/ts-essentials/blob/d0ad79f3e022f9ac97b5ec08314a20d3286db9dc/lib/paths/index.ts#L25-L42) would be the best place.

At the moment of reading, I have 2 concerns:

1. The name of the config option (i.e. `wildcard`) is too generic

When an engineer looks at `Paths<Company, { wildcard: '*' }>`, it's unclear what wildcard is used for:
- Will it use any object properties to find the matched path?
- Is it a specific shortcut to choose the first/last object property where the matched path exists?
- Is it something else?

I believe it will confuse engineers at first glance and may mislead them.

To help engineers, I suggest renaming `wildcard` to something else, e.g. `arrayIndexAccessor`, `arrayIndexWildcard`, etc. And then, for example, use `arrayIndexWildcard: '-1'` by default (it also keeps backwards compatibility) and you can override it to `arrayIndexWildcard: '*'` in your library.

What do you think?

2. Be cautious when setting the value to `*`

Although I agree that `'employees.*.name'` looks neater than `employees.-1.name`, without much context, it may still confuse engineers.

[...]

PR #412, opened 2024-10-01
feat(Paths): Add optional `anyArrayIndexAccessor`

mylesj:
## PR Checklist

- [x] Addresses an existing open issue: related to 
    - resolves: #411

## Overview

Adds optional configuration that allows wildcards to satisfy array index access.

### DIFF
--- a/lib/paths/index.ts
+++ b/lib/paths/index.ts
@@ -27,18 +27,25 @@ type RecursivePathsOptions = {
  * majority of use cases. If by any chance it doesn't fit you, feel free to
  * increase the value. However, this may increase the chance of getting
  * `Type instantiation is excessively deep and possibly infinite` error.
+ *
+ * @param anyArrayIndexAccessor By default there is no wildcard access to
+ * array indices - usage must be intentionally configured.
  */
 type DefaultPathsOptions = {
   depth: 7;
+  anyArrayIndexAccessor: `${number}`;
 };
 
 /**
  * @param depth This option restricts the depth of the paths lookup and removes `Type
  * instantiation is excessively deep and possibly infinite` errors for
  * potentially infinite types.
+ *
+ * @param anyArrayIndexAccessor This wildcard will satisfy any array index if defined.
  */
 type PathsOptions = {
   depth: number;
+  anyArrayIndexAccessor: string;
 };
 
 type Append<Tuple extends any[]> = [...Tuple, 0];
@@ -54,7 +61,7 @@ type RecursivePaths<
       ValueOf<{
         [Key in keyof Type]: Key extends Pathable
           ?
-              | `${Key}`
+              | `${AnyArrayIndexAccessorOrKey<Key, UserOptions>}`
               | (CallOptions["depth"]["length"] extends UserOptions["depth"]
                   ? // Stop at the configured depth
                     never
@@ -72,7 +79,7 @@ type RecursivePaths<
                         ? IsNever<Rest> extends true
                           ? never
                           : Rest extends Pathable
-                          ? `${Key}.${Rest}`
+                          ? `${AnyArrayIndexAccessorOrKey<Key, UserOptions>}.${Rest}`
                           : never
                         : never
                       : never
@@ -95,6 +102,10 @@ type UnsafePaths<Type, Options extends Required<PathsOptions>> = Type extends Ty
     ? RecursivePaths<Type, Options, DefaultRecursivePathsOptions>
     : never
   : never;
+
+type AnyArrayIndexAccessorOrKey<Key extends Pathable, UserOptions extends Required<PathsOptions>> = Key extends number
+  ? Key | UserOptions["anyArrayIndexAccessor"]
+  : Key;
 
 export type Paths<Type, OverridePathOptions extends Partial<PathsOptions> = {}> = UnsafePaths<
   Type,
--- a/test/paths.ts
+++ b/test/paths.ts
@@ -186,3 +186,31 @@ function testDepth2() {
     Assert<Assignable<P, "nested.tuple.2.good">>,
   ];
 }
+
+/**
+ * Configuration allows wildcard accessors for array indices.
+ */
+function testAnyArrayIndexAccessor() {
+  type With = Paths<ComplexNestedRequired, { anyArrayIndexAccessor: "*" }>;
+  type Without = Paths<ComplexNestedRequired>;
+
+  type cases_with_any_accessor = [
+    Assert<Has<With, "nested.array.*">>,
+    Assert<Has<With, "nested.array.*.bar">>,
+    Assert<Has<With, `nested.array.${number}`>>,
+    Assert<Has<With, `nested.array.${number}.bar`>>,
+    Assert<Has<With, "nested.tuple.0">>,
+    Assert<Has<With, "nested.tuple.2.good">>,
+    // @ts-expect-error: tuples should not be wildcard-able
+    Assert<Has<With, "nested.tuple.*">>,
+    // @ts-expect-error: tuples should not be wildcard-able
+    Assert<Has<With, "nested.tuple.*.good">>,
+  ];
+
+  type cases_without_any_accessor = [
+    // @ts-expect-error: anyArrayIndexAccessor is not configured
+    Assert<Has<Without, "nested.array.*">>,
+    // @ts-expect-error: anyArrayIndexAccessor is not configured
+    Assert<Has<Without, "nested.array.*.bar">>,
+  ];
+}
--- a/.changeset/slow-dolphins-fetch.md
+++ b/.changeset/slow-dolphins-fetch.md
@@ -1,0 +1,6 @@
+---
+"ts-essentials": patch
+---
+
+Add `Paths<Type, { anyArrayIndexAccessor: '*' }>`, a string literal representing a catch-all or "wildcard" when indexing
+on arrays.
--- a/lib/paths/README.md
+++ b/lib/paths/README.md
@@ -16,6 +16,16 @@ It supports:
 - Object types
 - Tuples
 - As a side effect of `number`, float number access for arrays
+- Wildcard array indices (opt-in)
+
+An optional symbol may be specified that will satisfy any array index
+
+```ts
+type WildcardCompanyPaths = Paths<Company, { anyArrayIndexAccessor: "*" }>;
+
+const anyEmployee: WildcardCompanyPaths = 'employees.*';
+const anyEmployeeName: WildcardCompanyPaths = 'employees.*.name';
+```
 
 ⚠️ Limitations:
 


test/paths.ts
```diff
@@ -186,3 +186,27 @@ function testDepth2() {
...
+  type Without = Paths<ComplexNestedRequired>;
+
+  type cases_with_any_accessor = [
+    Assert<Has<With, "nested.array.*">>,
+    Assert<Has<With, "nested.array.*.bar">>,
+    Assert<Has<With, `nested.array.${number}`>>,
+    Assert<Has<With, `nested.array.${number}.bar`>>,
+    Assert<Has<With, "nested.tuple.0">>,
+    Assert<Has<With, "nested.tuple.2.good">>,
+  ];
```
mylesj, line 204: I felt like adding this might be appropriate as I'm not sure that `nested.tuple.*` makes semantic sense. 

```typescript
    // @ts-expect-error: tuples should not be wildcard-able
    Assert<Has<With, "nested.tuple.*">>,
    // @ts-expect-error: tuples should not be wildcard-able
    Assert<Has<With, "nested.tuple.*.good">>,
```

However I could not get it working, trying several avenues of differentiating tuples from arrays - I must be misunderstanding something. This doesn't block my application but if you agree and would rather conservative exposure, I'm going to need some help with it.

tsconfig.test.json
```diff
@@ -1,4 +1,4 @@
 {
   "extends": "./tsconfig.json",
-  "include": ["lib/**/*.ts"]
-}
+  "include": ["lib/**/*.ts", "test/**/*.ts"]
```
mylesj, line 3: This is a drive-by fix - the tests are quite neat but they weren't actually running for me. I believe this was the intent?

lib/paths/index.ts
```diff
@@ -20,25 +21,33 @@ type DefaultRecursivePathsOptions = {
  */
 type RecursivePathsOptions = {
   depth: any[];
+  anyArrayIndexAccessor: string | false;
```
Beraliv, line 24: `RecursivePathsOptions` is required for `CallOptions`. Since you don't accumulate anything during recursive calls (unlike `depth` where type will contain the number of recursive calls), I suggest removing the value here

lib/paths/index.ts
```diff
@@ -54,7 +63,7 @@ type RecursivePaths<
       ValueOf<{
         [Key in keyof Type]: Key extends Pathable
           ?
-              | `${Key}`
+              | `${AnyArrayIndexAccessorOrKey<Key, CallOptions>}`
```
Beraliv, line 66: Since you don't update `anyArrayIndexAccessor` in `CallOptions` (unlike `depth`), I suggest using `UserOptions`:

```suggestion
              | `${AnyArrayIndexAccessorOrKey<Key, UserOptions>}`
```

lib/paths/index.ts
```diff
@@ -67,11 +76,12 @@ type RecursivePaths<
                           UserOptions,
                           {
                             depth: Append<CallOptions["depth"]>;
+                            anyArrayIndexAccessor: AnyArrayIndexAccessor<Value, UserOptions>;
```
Beraliv, line 79: As per comment above, `anyArrayIndexAccessor` is static, I suggest removing this line:

```suggestion
```

lib/paths/index.ts
```diff
@@ -94,6 +104,17 @@ type UnsafePaths<Type, Options extends Required<PathsOptions>> = Type extends Ty
     : never
   : never;
 
+type AnyArrayIndexAccessor<Type, Options extends Required<PathsOptions>> = Type extends any[]
```
Beraliv, line 107: You can safely remove type `AnyArrayIndexAccessor` since `anyArrayIndexAccessor` is always a string

lib/paths/index.ts
```diff
@@ -94,6 +104,17 @@ type UnsafePaths<Type, Options extends Required<PathsOptions>> = Type extends Ty
...
+  ? Options["anyArrayIndexAccessor"] extends string
+    ? `${Options["anyArrayIndexAccessor"]}`
+    : false
+  : false;
+
+type AnyArrayIndexAccessorOrKey<
+  Key extends Pathable,
+  Options extends RecursivePathsOptions,
+> = Options["anyArrayIndexAccessor"] extends false ? Key : Key | Options["anyArrayIndexAccessor"];
+
```
Beraliv, line 117: You can safely remove boolean conditional types since the config option is always a string.

test/paths.ts
Beraliv: Give `` `${number}` `` a try as I recommended above, it should help.

Tuples have specific keys: they are numeric literal types (e.g. 0, 1, 2). A conditional type with `DefaultPathsOptions["anyArrayIndexAccessor"]` and `` `${Key}` `` in `type AnyArrayIndexAccessorOrKey` should be sufficient to differentiate tuples and arrays

tsconfig.test.json
Beraliv: Good spot, I broke it a couple of commits ago! Thanks for fixing it!

test/paths.ts
```diff
@@ -186,3 +186,27 @@ function testDepth2() {
...
+ * Configuration allows wildcard accessors for array indices.
+ */
+function testAnyArrayIndexAccessor() {
+  type With = Paths<ComplexNestedRequired, { anyArrayIndexAccessor: "*" }>;
+  type Without = Paths<ComplexNestedRequired>;
+
+  type cases_with_any_accessor = [
+    Assert<Has<With, "nested.array.*">>,
+    Assert<Has<With, "nested.array.*.bar">>,
+    Assert<Has<With, `nested.array.${number}`>>,
```
Beraliv, line 200: During my testing, I found that `"nested.array.*"` will be acceptable, but not `` `nested.array.${number}` ``. I couldn't figure it out why this happens, so if you face it during your testing and cannot figure it out how to fix it, please leave it. It's a known bug

Beraliv (changes requested):
@mylesj Hey! Thank you so much for taking your time and raising this PR! Your PR looks good, I suggested multiple changes to align with my expectations. Let me know what you think

Beraliv:
@mylesj for your benefit, the testing strategy of ts-essentials is to run typechecks on multiple versions (between 4.5 and 5.5). Currently, 3 typechecks have failed on version 4.6-4.8, as GHA doesn't contain `continue-on-error: true`, it will stop running others. Kindly ping me when you're ready, I will accept re-running typechecks once again.

test/paths.ts
mylesj: Ah of course, this makes more sense.

I was trying a combination along the lines of this and it would either always swallow all arrays or all tuples:
```typescript
IsTuple<Type> extends Type 
  ? A 
  : Type extends any[]
  ? B
  : C
```

But I agree, the condition ends up much simpler when the `Key` is already inferred as `number`.

test/paths.ts
mylesj: Can you confirm if this is still an issue following the update and if so which version of TypeScript?

mylesj:
> _@mylesj for your benefit, the testing strategy of ts-essentials is to run typechecks on multiple versions (between 4.5 and 5.5). Currently, 3 typechecks have failed on version 4.6-4.8, as GHA doesn't contain continue-on-error: true, it will stop running others. Kindly ping me when you're ready, I will accept re-running typechecks once again._

@Beraliv So I had a look at the tests under the different versions and I'm reasonably sure this change set isn't directly related - more a consequence of my re-enabling the tests and surfacing existing errors on `master`. Specifically: 
  - `test/paths.ts` fail on **4.6**
  - `test/xor.ts` fail on: **4.6**, **4.7** & **4.8**

That said, I spent some time trying to understand the paths issue but can't for the moment see what the problem might be - something is truncating the paths prematurely or preventing recursion. I'd ask that you take a closer look to confirm whether this is related to recent changes.

If you agree, I've selectively disabled those specific test suites for now. Otherwise I'm going to be unavailable for the next week or so but will take another look when I'm back.

test/paths.ts
Beraliv: I'll double-check it on the weekend

lib/paths/index.ts
```diff
@@ -78,7 +80,9 @@ type RecursivePaths<
                         > extends infer Rest
                         ? IsNever<Rest> extends true
                           ? never
-                          : `${AnyArrayIndexAccessorOrKey<Key, UserOptions>}.${Rest & string}`
+                          : Rest extends ReallyPrimitive
+                          ? `${AnyArrayIndexAccessorOrKey<Key, UserOptions>}.${Rest}`
+                          : never
```
mylesj, line 85: @Beraliv 

I think I got it, this `Rest & string` is broadening the path to any `anything.nested.${string}` in earlier versions.

test/paths.ts
Beraliv: When I enabled tests in another branch (which is created from `master` branch), I could reproduce the problem: commit https://github.com/ts-essentials/ts-essentials/pull/413/commits/64fffdc70bf9a5e86482fadb894c8604e7baa92b

Screenshot 2024-10-07 at 09 31 25

The problem existed in TypeScript versions 4.5-4.8, it's fixed after this commit - https://github.com/ts-essentials/ts-essentials/pull/413/commits/98258d5fbccd4e6d0ba4d3e32ec58ce42c9f33ab

I'll get it merged today, please rebase it and you won't need to make any other changes

lib/paths/index.ts
```diff
@@ -69,9 +77,9 @@ type RecursivePaths<
                             depth: Append<CallOptions["depth"]>;
                           }
                         > extends infer Rest
-                        ? IsNever<Rest> extends true
-                          ? never
-                          : `${Key}.${Rest & string}`
+                        ? Rest extends Exclude<Primitive, symbol>
```
Beraliv, line 80: You may look for `Pathable` here:

```suggestion
                        ? Rest extends Pathable
```

I've tested it in https://github.com/ts-essentials/ts-essentials/pull/413 (commit https://github.com/ts-essentials/ts-essentials/pull/413/commits/98258d5fbccd4e6d0ba4d3e32ec58ce42c9f33ab) and it fixed the problem for TypeScript versions 4.5-4.8

tsconfig.test.json
```diff
@@ -1,4 +1,8 @@
 {
   "extends": "./tsconfig.json",
-  "include": ["lib/**/*.ts"]
-}
+  "include": ["lib/**/*.ts", "test/**/*.ts"],
+  "exclude": [
+    // TODO: fix 4.6.4; 4.7.4; 4.8.4
+    "test/xor.ts",
+  ]
```
Beraliv, line 7: Please remove it when https://github.com/ts-essentials/ts-essentials/pull/413 is merged

Beraliv (commented):
3 more comments, otherwise LGTM!

mylesj:
Fantastic. So I'm AFK / on vacation this week but will wrap things up at the weekend.

tsconfig.test.json
Beraliv: The PR is just merged ✅

lib/paths/index.ts
```diff
@@ -96,6 +103,12 @@ type UnsafePaths<Type, Options extends Required<PathsOptions>> = Type extends Ty
     : never
   : never;
 
+// prettier-ignore
```
Beraliv, line 106: Is it still required?

mylesj: No, I just thought it read better - the line length seems to be quite long. I'll remove it though.

Beraliv (approved):
Great work! Thank you for your efforts, LGTM now 👏

Beraliv (commented):
Last thing: please add a changeset (part of contributing guide): you have to run `yarn changeset` and add a minor change with your preferred message

mylesj:
Ah my bad, I'm not familiar with this approach to change log. Done - thanks for sticking with me, appreciate the support!

Beraliv:
Great, thank you @mylesj for your efforts! Changes look good, so I will merge it now

PR #414, opened 2024-10-21
Fix: Behaviour of `OptionalKeys` when instantiated with primitives and arrays

som-sm:
## PR Checklist

## Overview

Currently, the `OptionalKeys` type doesn’t handle primitives and arrays as expected.

### Handling Primitives Correctly
Since `OptionalKeys` is a [homomorphic mapped type](https://github.com/microsoft/TypeScript/pull/26063), so when it’s instantiated with a primitive `Type`, it simply returns back `Type`. As a result, instantiating `OptionalKeys` with `string` yields `string[keyof string]`, which isn’t the intended result.

The fix here is straightforward, we simply bypass primitives by changing the conditional from `Type extends unknown` to `Type extends object`. So, when `OptionalKeys` is instantiated with primitives, the condition evaluates to false, returning `never` as expected.

### Handling Arrays Correctly
Ideally, `OptionalKeys<[1, 2?]>` should return `"1"`, but instead, it returns a bunch of unintended things.

Here's the existing code for reference:
```ts
export type OptionalKeys<Type> = Type extends unknown
  ? {
      [Key in keyof Type]-?: undefined extends { [Key2 in keyof Type]: Key2 }[Key] ? Key : never;
    }[keyof Type]
  : never;
```
Let's break down what happens when `OptionalKeys` is instantiated with `[1, 2?]`.

- Since the inner mapped type (`{ [Key2 in keyof Type]: Key2 }`) is homomorphic, it **correctly** preserves the tuple structure and evaluates to `[Key2, Key2?]` (Note: `Key2` is insignificant here; could have simply been `never` as well). Therefore, the conditional `undefined extends ...` evaluates to true only when `Key` is `"1"`.
- Consequently, because the outer mapped type is also homomorphic, it **correctly** evaluates to `[never, "1"]`.
- However, the problem arises when we try to do `[never, "1"][keyof T]` as this yields a bunch of unintended keys. Instead, if we index it with the `number` type, we get the desired result of "1".

The fix involves using `number` as the index for arrays instead of `keyof Type`.

### Additional Improvements & Tests:
1. Added a couple of more test cases to cover arrays and functions.
2. Replaced `Key2` with `never` in the inner mapped type for better readability. Using a specific type like `Key2` can imply that its value is being utilised, which isn’t the case here.

### DIFF
--- a/.changeset/ninety-forks-poke.md
+++ b/.changeset/ninety-forks-poke.md
@@ -2,4 +2,4 @@
 "ts-essentials": patch
 ---
 
-Fix `OptionalKeys<Type>`, it now works as expected with primitives and arrays
+`OptionalKeys<Type>` returns `never` for primitives and returns only optional indices for arrays and tuples
--- a/.changeset/ninety-forks-poke.md
+++ b/.changeset/ninety-forks-poke.md
@@ -1,0 +1,5 @@
+---
+"ts-essentials": patch
+---
+
+Fix `OptionalKeys<Type>`, it now works as expected with primitives and arrays
--- a/lib/optional-keys/index.ts
+++ b/lib/optional-keys/index.ts
@@ -1,5 +1,5 @@
 export type OptionalKeys<Type> = Type extends unknown
   ? {
-      [Key in Exclude<keyof Type, never>]-?: undefined extends { [Key2 in keyof Type]: Key2 }[Key] ? Key : never;
+      [Key in Exclude<keyof Type, never>]-?: undefined extends { [Key2 in keyof Type]: never }[Key] ? Key : never;
     }[keyof Type]
   : never;
--- a/lib/optional-keys/index.ts
+++ b/lib/optional-keys/index.ts
@@ -1,5 +1,5 @@
 export type OptionalKeys<Type> = Type extends unknown
   ? {
-      [Key in keyof Type]-?: undefined extends { [Key2 in keyof Type]: Key2 }[Key] ? Key : never;
+      [Key in Exclude<keyof Type, never>]-?: undefined extends { [Key2 in keyof Type]: Key2 }[Key] ? Key : never;
     }[keyof Type]
   : never;
--- a/lib/optional-keys/index.ts
+++ b/lib/optional-keys/index.ts
@@ -1,5 +1,5 @@
 export type OptionalKeys<Type> = Type extends object
-  ? {
-      [Key in keyof Type]-?: Type extends Required<Pick<Type, Key>> ? never : Key;
-    }[keyof Type & (Type extends ReadonlyArray<any> ? number : keyof Type)]
+  ? keyof {
+      [Key in keyof Type as Type extends Required<Pick<Type, Key>> ? never : Key]: any;
+    }
   : never;
--- a/lib/optional-keys/index.ts
+++ b/lib/optional-keys/index.ts
@@ -1,5 +1,5 @@
 export type OptionalKeys<Type> = Type extends object
   ? {
-      [Key in keyof Type]-?: undefined extends { [Key2 in keyof Type]: never }[Key] ? Key : never;
+      [Key in keyof Type]-?: Type extends Required<Pick<Type, Key>> ? never : Key;
     }[keyof Type & (Type extends ReadonlyArray<any> ? number : keyof Type)]
   : never;
--- a/lib/optional-keys/index.ts
+++ b/lib/optional-keys/index.ts
@@ -1,5 +1,5 @@
-export type OptionalKeys<Type> = Type extends unknown
+export type OptionalKeys<Type> = Type extends object
   ? {
-      [Key in Exclude<keyof Type, never>]-?: undefined extends { [Key2 in keyof Type]: never }[Key] ? Key : never;
-    }[keyof Type]
+      [Key in keyof Type]-?: undefined extends { [Key2 in keyof Type]: never }[Key] ? Key : never;
+    }[keyof Type & (Type extends ReadonlyArray<any> ? number : keyof Type)]
   : never;
--- a/test/index.ts
+++ b/test/index.ts
@@ -457,22 +457,23 @@ function testPickProperties() {
 
 function testOptionalKeys() {
   type cases = [
-    // @ts-expect-error converts to Number and gets its optional keys
     Assert<IsExact<OptionalKeys<number>, never>>,
-    // @ts-expect-error converts to String and gets its optional keys
     Assert<IsExact<OptionalKeys<string>, never>>,
-    // wtf?
-    Assert<IsExact<OptionalKeys<boolean>, () => boolean>>,
-    // @ts-expect-error converts to BigInt and gets its optional keys
+    Assert<IsExact<OptionalKeys<boolean>, never>>,
     Assert<IsExact<OptionalKeys<bigint>, never>>,
-    // wtf?
-    Assert<IsExact<OptionalKeys<symbol>, string | ((hint: string) => symbol) | (() => string) | (() => symbol)>>,
+    Assert<IsExact<OptionalKeys<symbol>, never>>,
     Assert<IsExact<OptionalKeys<undefined>, never>>,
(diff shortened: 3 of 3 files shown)


Beraliv:
Hey @som-sm! Thank you for taking your time and efforts to contribute to ts-essentials! I won't be able to have a look at your changes today, so I will come back to you tomorrow! Thank you for understanding in advance

Beraliv:
Quickly looking at the code, your changes look reasonable to me. I let the CI run on all TS versions to see if there are any regressions. I'll double-check the changes more thoroughly today to see if I missed anything. I'll ping you if needed

som-sm:
Looks like something probably got fixed in `v4.9`.

From a quick test, the issue seems to be here:
```ts
export type Test<Type> = {
  [Key in keyof Type]-?: [{ [Key2 in keyof Type]: never }[Key]];
};

type T = Test<[1, 2?]>;
```
For v4.8 and below
`T` is of type `[[never], [never]]`

Which is why the conditional in `OptionalKeys` is not becoming true for the second item.

Whereas for v4.9 and above
`T` is of type `[[never], [undefined]]`

som-sm:
I've refactored the way we check for optional properties, this seems to be working fine for v4.8.

```diff
export type OptionalKeys<Type> = Type extends object
  ? {
-     [Key in keyof Type]-?: undefined extends { [Key2 in keyof Type]: Key2 }[Key] ? Key : never;
+     [Key in keyof Type]-?: Type extends Required<Pick<Type, Key>> ? never : Key;
    }[keyof Type & (Type extends ReadonlyArray<any> ? number : keyof Type)]
  : never;
```

test/index.ts
```diff
@@ -457,22 +457,23 @@ function testPickProperties() {
...
     Assert<IsExact<OptionalKeys<Function>, never>>,
     Assert<IsExact<OptionalKeys<Date>, never>>,
     Assert<IsExact<OptionalKeys<Error>, "stack">>,
     Assert<IsExact<OptionalKeys<RegExp>, never>>,
+    Assert<IsExact<OptionalKeys<() => void>, never>>,
+    Assert<IsExact<OptionalKeys<any[]>, never>>,
+    Assert<IsExact<OptionalKeys<[1, 2]>, never>>,
+    Assert<IsExact<OptionalKeys<[1, 2?]>, "1">>,
+    Assert<IsExact<OptionalKeys<{ (): void; a?: 1 }>, "a">>,
+    Assert<IsExact<OptionalKeys<{ (): void; a: 1 }>, never>>,
```
Beraliv, line 476: I believe you're testing the same thing as on lines L478-479?

test/index.ts
```diff
@@ -492,16 +493,22 @@ function testOptionalKeys() {
...
     Assert<IsExact<RequiredKeys<null>, never>>,
     Assert<IsExact<RequiredKeys<Function>, keyof Function>>,
     Assert<IsExact<RequiredKeys<Date>, keyof Date>>,
     Assert<IsExact<RequiredKeys<Error>, "name" | "message">>,
     Assert<IsExact<RequiredKeys<RegExp>, keyof RegExp>>,
+    Assert<IsExact<RequiredKeys<() => void>, never>>,
+    Assert<IsExact<RequiredKeys<any[]>, keyof any[]>>,
+    Assert<IsExact<RequiredKeys<[1, 2]>, keyof [1, 2]>>,
+    Assert<IsExact<RequiredKeys<{ (): void; a?: 1 }>, never>>,
+    Assert<IsExact<RequiredKeys<{ (): void; a: 1 }>, "a">>,
```
Beraliv, line 510: Same question here: are you testing the same thing as on lines L513-514?

.changeset/ninety-forks-poke.md
```diff
@@ -0,0 +1,5 @@
+---
+"ts-essentials": patch
+---
+
+Fix `OptionalKeys<Type>`, it now works as expected with primitives and arrays
```
Beraliv, line 5: I suggest being more specific on what was changed:

```suggestion
`OptionalKeys<Type>` returns `never` for primitives and returns only optional indices for arrays and tuples
```

Beraliv:
@som-sm thank you for your patience! I've reviewed your changes in detail and left 3 comments. Otherwise, it LGTM! Please address my comments and I'm happy to merge your changes.

test/index.ts
som-sm: Not exactly, L475-476 tests **functions** with optional properties (they have a call signature `(): void`), whereas L478-479 tests just plain objects.

test/index.ts
som-sm: Same as above, L509 and L510 test functions with optional and required properties, respectively.

.changeset/ninety-forks-poke.md
som-sm: Agreed!
Updated the PR.

som-sm:
@Beraliv Thanks for reviewing the PR. I've replied back to all the comments, please have a look.

Also, there's scope for similar improvements in `ReadonlyKeys` and `WritableKeys` types, I'll raise those PRs soon.

som-sm:
@Beraliv, while fixing the `ReadonlyKeys` type, I realised there's a **much** simpler way of implementing the `OptionalKeys` type that **doesn't** require any special handling for arrays. I've updated the PR, please refer to [this commit](https://github.com/ts-essentials/ts-essentials/pull/414/commits/e841cd4a659916fbbd1bf302a2dd004b6297c2a5).

test/index.ts
Beraliv: Could you explain why it's  important? If a function returns `never` (an assertion above), there is no difference between object with / without function?

lib/optional-keys/index.ts
```diff
@@ -1,5 +1,5 @@
-export type OptionalKeys<Type> = Type extends unknown
-  ? {
-      [Key in keyof Type]-?: undefined extends { [Key2 in keyof Type]: Key2 }[Key] ? Key : never;
-    }[keyof Type]
+export type OptionalKeys<Type> = Type extends object
+  ? keyof {
+      [Key in keyof Type as Type extends Required<Pick<Type, Key>> ? never : Key]: any;
```
Beraliv, line 3: (Optional) Anything can be a value, you can simply use a numeric literal as the shortest option (e.g. 1) to avoid usage of `any`

Beraliv (commented):
Nice usage of key remapping, overall LGTM, but a follow-up on tests

test/index.ts
som-sm: For example, if a `Function` check is added that directly returns `never`, **only** the test case with object containing **both** a call signature and an optional property (L475) would fail, ensuring such changes are blocked.

```ts
export type OptionalKeys<Type extends object> = Type extends Function
  ? never
  : keyof {
      [Key in keyof Type as Type extends Required<Pick<Type, Key>> ? never : Key]: any;
    };
```

lib/optional-keys/index.ts
som-sm: IMO, it's better to have `any` just to signify that the value can be anything. Having something more specific, like `1`, might not convey the same meaning and could create confusion.

Can change it to `unknown` though, if `any`s are to be avoided.

Beraliv: There are no best practices in this repo (yet), so `any` works as well. I'll leave it to you, feel free to resolve the comment if you'd like to stay with `any`

som-sm: Let's leave it as `any` then.

test/index.ts
Beraliv: It makes sense to me, thank you for your explanation! Although I believe the risk of regression with functions is low, I'm happy to leave these test cases

Beraliv:
@som-sm I'm happy to merge your PR now, let me know what you think about my last comment with `any`

som-sm:
@Beraliv I think it's just fine to keep `any` as the value there.

Beraliv:
@som-sm no objections from me! Great work, merging your PR now!

PR #416, opened 2024-10-29
Fix: `ReadonlyKeys` and `WritableKeys` types with unions, arrays and index signatures

som-sm:
## PR Checklist

## Overview

**This PR:**
1. Fixes the implementation of `ReadonlyKeys` and `WritableKeys` to ensure correct results with unions, arrays, and index signatures, following the updates made in [PR#414](https://github.com/ts-essentials/ts-essentials/pull/414).
2. Adds the `semver` package to generalise the management of conditional test files.
3. Adjusts the regular expressions in `find-ts-version.js` to capture the TS version more accurately.

### Why conditional test cases?
Unfortunately, some test cases do not pass in older TS versions (i.e., <= v5.2) due to the unexpected behaviour of `Pick` in these versions. Here are a few examples:

- When extracting the index type from an index signature, `Pick` fails to preserve the `readonly` property modifier:
```ts
type T1 = Pick<{ readonly [x: string]: 1 }, string>
//   ^? type T1 = { [x: string]: 1; }
```

- Similarly, when extracting the `number` type from arrays and tuples, `Pick` again fails to preserve the `readonly` property modifier:
```ts
type T2 = Pick<readonly boolean[], number>
//   ^? type T2 = { [x: number]: boolean; }

type T3 = Pick<readonly [1, 2], number>
//   ^? type T3 = { [x: number]: 1 | 2; }
```

Playground link: https://tsplay.dev/NdxxQN

Due to this inconsistent behaviour of `Pick`, the `ReadonlyKeys` and `WritableKeys` types exhibit slight variations in their results for TS versions above and below v5.3, which is why we need conditional test files.

image

<br />
<br />

**NOTE:** Both these behaviours are still considerably better than the existing behaviour.


Beraliv:
@som-sm you're awesome! Thanks in advance for your contributions 🧡 I'll come back to you later today or tomorrow and review your changes in-depth

scripts/find-ts-version.js
```diff
@@ -1,8 +1,8 @@
 // @ts-check
 const fs = require("fs");
 
-const FIRST_TYPESCRIPT_LINE = /typescript\@\^?(\d+.\d+(.\d+)?):/;
-const TYPESCRIPT_VERSION_LINE = /version "(\d+.\d+)(.\d+)?"/;
+const FIRST_TYPESCRIPT_LINE = /typescript@\S*:/;
```
Beraliv, line 4: I see that it may be useful for beta and preview versions of TypeScript. But currently, we don't use those, so it's unclear why this change is important. Would you be able to clarify your reasons?

scripts/find-ts-version.js
```diff
@@ -1,8 +1,8 @@
...
+const TYPESCRIPT_VERSION_LINE = /version "(\d+\.\d+\.\d+)"/;
```
Beraliv, line 5: Originally, `TYPESCRIPT_VERSION_LINE` was used to generate `test/ts-version.ts` with `type TsVersion` which was used in unit tests, when output was different for different TypeScript versions.

I've tidied up the usage just now in https://github.com/ts-essentials/ts-essentials/commit/25b538c27d8917f798a3a87d7003102a32f1d412

But it's still useful in e.g. https://github.com/ts-essentials/ts-essentials/blob/master/test/strict-extract.ts#L53

I still believe that writing `TsVersion extends "4.5"` is easier than `TsVersion extends "4.5.5"`, but to be honest, there is no much difference.

Therefore I suggest:

1. Please leave a JSDoc comment for `TYPESCRIPT_VERSION_LINE` to explain that, currently, there are 2 purposes for this Regexp: it defines `TsVersion` and it is used in conditional tests for a `semver` library

2. Please update `strict-extract.ts` unit test so the CI check is passed for TypeScript 4.5

What do you think?

scripts/update-test-tsconfig.js
```diff
@@ -12,18 +15,17 @@ const CONDITIONAL_FILES = {
...
   let exclude = (json.exclude || []).filter((v) => !ALL_CONDITIONAL_FILES.includes(v));
 
-  if (["4.5", "4.6", "4.7", "4.8"].includes(tsVersion)) {
-    // XOR utility type never inferred the right types in functions parameters
-    // for TypeScript versions between 4.5.x and 4.8.x
-
-    exclude.push(CONDITIONAL_FILES.XOR_FUNCTION_PARAMETERS);
-  }
+  Object.entries(CONDITIONAL_FILES).forEach(([file, version]) => {
+    if (semver.satisfies(tsVersion, version)) {
```
Beraliv, line 25: I suggest renaming a `version` to `tsVersionConstraint`, because it may contain multiple versions, a range, etc

```suggestion
  Object.entries(CONDITIONAL_FILES).forEach(([file, tsVersionConstraint]) => {
    if (semver.satisfies(tsVersion, tsVersionConstraint)) {
```

WDYT?

test/index.ts
```diff
@@ -557,23 +557,55 @@ function testNonEmptyArray() {
...
-  type Test = Assert<IsExact<Actual, Expected>>;
+  type cases = [
+    // @ts-expect-error primitives not allowed
+    Assert<IsExact<ReadonlyKeys<string>, keyof string>>,
+    // @ts-expect-error primitives not allowed
+    Assert<IsExact<ReadonlyKeys<number>, keyof number>>,
+    // @ts-expect-error primitives not allowed
+    Assert<IsExact<ReadonlyKeys<boolean>, keyof boolean>>,
+    // @ts-expect-error primitives not allowed
+    Assert<IsExact<ReadonlyKeys<bigint>, keyof bigint>>,
```
Beraliv, line 568: (optional) I've checked first and last supported version and it's likely to be `never`:

```suggestion
    Assert<IsExact<ReadonlyKeys<bigint>, never>>,
```

test/index.ts
```diff
@@ -557,23 +557,55 @@ function testNonEmptyArray() {
...
+    // @ts-expect-error primitives not allowed
+    Assert<IsExact<ReadonlyKeys<symbol>, keyof symbol>>,
+    // @ts-expect-error primitives not allowed
+    Assert<IsExact<ReadonlyKeys<null>, keyof null>>,
+    // @ts-expect-error primitives not allowed
+    Assert<IsExact<ReadonlyKeys<undefined>, keyof undefined>>,
```
Beraliv, line 574: (optional) `null` and `undefined` will definitely have no keys for first and last supported versions, therefore you can easily replace them with `never`:

```suggestion
    // @ts-expect-error primitives not allowed
    Assert<IsExact<ReadonlyKeys<null>, never>>,
    // @ts-expect-error primitives not allowed
    Assert<IsExact<ReadonlyKeys<undefined>, never>>,
```

test/index.ts
```diff
@@ -557,23 +557,55 @@ function testNonEmptyArray() {
...
 function testReadonlyKeys() {
-  type T = { readonly a: number; b: string };
-
-  type Actual = ReadonlyKeys<T>;
-
-  type Expected = "a";
-
-  type Test = Assert<IsExact<Actual, Expected>>;
+  type cases = [
+    // @ts-expect-error primitives not allowed
```
Beraliv, line 561: The description doesn't make much sense, because you still return something (and it's not `never`), therefore it should be updated for the majority of the primitives?

test/index.ts
Beraliv, line 574: you don't need `Assert` for this test cases since then will throw a TypeError, therefore I suggest removing `Assert` and expected values:

```suggestion
    // @ts-expect-error primitives not allowed
   ReadonlyKeys<string>,
    // @ts-expect-error primitives not allowed
    ReadonlyKeys<number>,
    // @ts-expect-error primitives not allowed
    ReadonlyKeys<boolean>,
    // @ts-expect-error primitives not allowed
    ReadonlyKeys<bigint>,
    // @ts-expect-error primitives not allowed
    ReadonlyKeys<symbol>,
    // @ts-expect-error primitives not allowed
    ReadonlyKeys<null>,
    // @ts-expect-error primitives not allowed
    ReadonlyKeys<undefined>,
```

test/writable-keys-arrays-and-index-signatures.ts
```diff
@@ -0,0 +1,22 @@
+import { AssertTrue as Assert, IsExact } from "conditional-type-checks";
+import { WritableKeys } from "../lib";
+
+function testWritableKeys() {
+  type cases = [
```
Beraliv, line 5: To avoid regression, please add the same list of test cases with primitives (without `Assert` and expected values)

Beraliv:
@som-sm I left my thoughts, please review my changes and let me know if you have any questions

scripts/find-ts-version.js
som-sm: 1. Yeah sure, I'll add a comment there.
2. Apologies, but I’m having trouble understanding the need for a conditional test in `strict-extract`. Because both test cases seem to pass in `v4.5` and later versions (playground link: https://tsplay.dev/NV8BMW). So, maybe we can remove the usage of `TsVersion` from `strict-extract` as well.

And maybe in the future we can switch to this if we want to avoid specifying the patch number:
```ts
TsVersion extends `4.5.${number}`
```

scripts/update-test-tsconfig.js
som-sm: Yeah, makes sense!

test/index.ts
som-sm: Yup, apologies I missed that!

@Beraliv Also, we currently have the `object` constraint in `ReadonlyKeys` and `WritableKeys`, but not in `OptionalKeys` and `RequiredKeys`. I feel we should add the constraint to the latter two types as well. WDYT?

scripts/find-ts-version.js
som-sm: The current regex only matches versions prefixed with `^`, but when testing locally, I often change the version to something like `~4.5` to get the latest patch of a minor version. However, this causes `yarn test` to fail due to the regex.

Also, since we're not using any of the captured groups from this regex, it felt like the regex could be simplified. Let me know if you think this approach might produce false positives. Maybe I should update the regex to this:

```suggestion
const FIRST_TYPESCRIPT_LINE = /^typescript@\S*:$/;
```

som-sm:
@Beraliv I've replied back to all your questions, and pushed all suggestions changes. Also, I've added the changeset, LMK if it reads fine.

Beraliv:
Thank you, I'll revisit the changes today

test/index.ts
Beraliv: Yeah, I think it should be aligned, but at the same time it introduces a breaking change. For the time being, please don't align it. Next time, we're making a major release, I'm happy to update it

scripts/find-ts-version.js
Beraliv: FYI Your link points to TS 5.0, you need 4.5 - https://tsplay.dev/wOQ9Rm (but it worked).

The issue exists in TS versions:

- 4.6 - https://tsplay.dev/mLVP2w
- 4.7 - https://tsplay.dev/w1DvAW

You will see:

```
Type '{ type: "dog" | "cat"; }' is not assignable to type 'Partial<Mouse>'.
  Types of property 'type' are incompatible.
    Type '"dog" | "cat"' is not assignable to type '"mouse" | undefined'.
      Type '"dog"' is not assignable to type '"mouse"'
```

Example - https://github.com/ts-essentials/ts-essentials/actions/runs/11626949786/job/32379475714

It depends on how `Partial` is implemented, it seems to be broken in versions 4.6 and 4.7

Anyway, I've updated the `StrictExtract` tests in https://github.com/ts-essentials/ts-essentials/commit/dc1cd6979dde65b440ba5e618ca136c5d45642a2

scripts/find-ts-version.js
Beraliv: @som-sm good point on `~4.5`, I'm happy to keep it

Beraliv:
@som-sm no new comments, responded to your questions, please have a look

Beraliv:
@all-contributors please add @som-sm

scripts/find-ts-version.js
som-sm: Interesting, doesn’t seem to be an issue with `Partial` but rather with how the type constraint is being enforced. Because the following fails `v4.5` and `v4.6`, but works in others.

Playground: https://tsplay.dev/WKQDpN
image

Anyways, I've pulled in your changes and updated the `strict-extract` conditional tests ([commit](https://github.com/ts-essentials/ts-essentials/pull/416/commits/3d54e93416812f155090dcb2528d53cbe13f09d6)). LGTM now!

test/index.ts
som-sm: Cool!!

Also, there's scope for adding constraints to type parameters in several other types as well, like the `Mark*` family of types. It may already be in the plan, but if not, we should probably think about this during the next major release.

som-sm:
@Beraliv I've synced with your changes & fixed the conditional tests for `strict-extract`. PR looks good to be merged!

test/index.ts
Beraliv: I'll create an issue where we can discuss all major changes. For now, I'm going to release patch changes only

scripts/find-ts-version.js
Beraliv: Great, thank you!

Beraliv (approved):
@som-sm thank you for all you efforts once again, the changes LGTM, so I'm going to merge them now

Beraliv:
Hey @som-sm! Just an FYI, today I updated the list of contributors and [added you to the list](https://github.com/ts-essentials/ts-essentials?tab=readme-ov-file#contributors), thank you for your multiple efforts once again!

som-sm:
Thanks @Beraliv, would love to actively contribute to this project. Feel free to tag me wherever you feel I can add value.

PR #417, opened 2024-11-03
Fix: Failing `test:fix` script

som-sm:
## PR Checklist

## Overview

@Beraliv Here's a tiny PR.

Just realised that the `test:fix` script now fails because it doesn’t update `tsconfig.test.json`. This update is now required since even the locked TypeScript version (i.e., `4.9.5`) needs a few files to be excluded due to the changes made in [this PR](https://github.com/ts-essentials/ts-essentials/pull/416).

I've removed running tests from the `test:fix` script in this PR. However, if you feel we should not remove it, then we could do it like this:
```diff
- "test:fix": "prettier --write {bench,lib,test}/**/*.ts && tsc --project tsconfig.test.json --outDir ./testDist && rimraf ./testDist"
+ "test:fix": "prettier --write {bench,lib,test}/**/*.ts && yarn test"
```

<br />

And either ways, I think we should probably rename the script to `prettier:fix` because the script doesn't really fix any tests. WDYT?

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,7 @@
     "formatDeclarations": "prettier --ignore-path *.js --write dist/*.d.ts",
     "prepublishOnly": "yarn test && yarn build && yarn formatDeclarations",
     "test": "node scripts/update-test-tsconfig.js && prettier -c {bench,lib,test}/**/*.ts && tsc --project tsconfig.test.json --outDir ./testDist && rimraf ./testDist",
-    "test:fix": "prettier --write {bench,lib,test}/**/*.ts",
+    "test:fix": "node scripts/update-test-tsconfig.js && prettier --write {bench,lib,test}/**/*.ts",
     "perf": "tsx ./bench/index.ts --benchErrorOnThresholdExceeded --benchPercentThreshold 1",
     "prerelease": "yarn test",
     "release": "yarn build && yarn formatDeclarations && yarn changeset publish",
--- a/package.json
+++ b/package.json
@@ -13,7 +13,7 @@
     "formatDeclarations": "prettier --ignore-path *.js --write dist/*.d.ts",
     "prepublishOnly": "yarn test && yarn build && yarn formatDeclarations",
     "test": "node scripts/update-test-tsconfig.js && prettier -c {bench,lib,test}/**/*.ts && tsc --project tsconfig.test.json --outDir ./testDist && rimraf ./testDist",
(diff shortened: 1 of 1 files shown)


Beraliv:
@som-sm thank you for contributing again! I'll re-visit your PRs later this week

som-sm:
@Beraliv Cool, no problem, please take your time.

Beraliv:
Hey @som-sm, I'm back! I'm reviewing your PR right now

package.json
```diff
@@ -13,7 +13,7 @@
     "formatDeclarations": "prettier --ignore-path *.js --write dist/*.d.ts",
     "prepublishOnly": "yarn test && yarn build && yarn formatDeclarations",
     "test": "node scripts/update-test-tsconfig.js && prettier -c {bench,lib,test}/**/*.ts && tsc --project tsconfig.test.json --outDir ./testDist && rimraf ./testDist",
-    "test:fix": "prettier --write {bench,lib,test}/**/*.ts && tsc --project tsconfig.test.json --outDir ./testDist && rimraf ./testDist",
+    "test:fix": "prettier --write {bench,lib,test}/**/*.ts",
```
Beraliv, line 16: Because `test:fix` script is expected to do more than `test` script, it makes sense to still run `tsc` for type-checking. Therefore I suggest returning your change and adding `node scripts/update-test-tsconfig.js &&` in front of `test:fix` script.

WDYT?

som-sm: @Beraliv You mean the only difference b/w `test` and `test:fix` should be that the former checks prettier formatting and the latter fixes prettier formatting? If so, what's the benefit of type checking within `test:fix`, is it so that contributors can just run a single script that does everything?

som-sm: Cool, I've added `node scripts/update-test-tsconfig.js` to the script in [this commit](https://github.com/ts-essentials/ts-essentials/pull/417/commits/90ee600486bc76d4e0683f8b5bf33f0f1ad5f4e8).

som-sm: @Beraliv I think we should also update the default `tsconfig.test.json` because now every time we run `test:fix` it would show a change in that file. WDYT?

image

Beraliv: Yeah, I already saw it, added it to the main branch in https://github.com/ts-essentials/ts-essentials/commit/bc00a4f773967b4d23a37cee753c9cce3ef72b90

Beraliv:
Great, it LGTM now! I'm merging it now

PR #421, opened 2024-11-19
Fix: `MarkRequired` and `MarkWritable` types when `Keys` is `any`

som-sm:
## PR Checklist

## Overview

The `MarkRequired` and `MarkWritable` types currently have issues when the second type argument is `any`. This is because their implementation relies on `Pick`, which does not handle `any` correctly.

For eg, `MarkRequired<{ a?: 1; b?: 2 }, any>` currently evaluates to type `{ [x: string]: any; a?: 1; b?: 2 }`, instead it should have simply evaluated to type `{ a: 1; b: 2 }`.

This PR refactors these types to use a combination of `Omit` and `Exclude` instead of `Pick`, ensuring proper functionality with `any`.

### DIFF
--- a/.changeset/orange-cycles-wink.md
+++ b/.changeset/orange-cycles-wink.md
@@ -1,0 +1,5 @@
+---
+"ts-essentials": patch
+---
+
+Fix `MarkRequired<Type, Keys>` & `MarkWritable<Type, Keys>` types when `Keys` is `any`
--- a/lib/mark-required/index.ts
+++ b/lib/mark-required/index.ts
@@ -1,5 +1,5 @@
 import { Prettify } from "../prettify";
 
 export type MarkRequired<Type, Keys extends keyof Type> = Type extends Type
-  ? Prettify<Type & Required<Pick<Type, Keys>>>
+  ? Prettify<Type & Required<Omit<Type, Exclude<keyof Type, Keys>>>>
   : never;
--- a/lib/mark-writable/index.ts
+++ b/lib/mark-writable/index.ts
@@ -3,5 +3,5 @@ import { WritableKeys } from "../writable-keys";
 import { Prettify } from "../prettify";
 
 export type MarkWritable<Type, Keys extends keyof Type> = Type extends Type
-  ? Prettify<Readonly<Type> & Writable<Pick<Type, (Type extends object ? WritableKeys<Type> : never) | Keys>>>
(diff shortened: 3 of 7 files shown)


Beraliv:
Hey @som-sm! Thank you for a PR, I'll have a look later this week

Beraliv:
@som-sm apologies for taking it longer than I expected. I have more time this week, so I'll check it this week

som-sm:
@Beraliv No worries, please take your time.

Beraliv:
@som-sm Sorry, it takes more time to understand the use case. The reasoning isn't clear: I would basically do the opposite and prohibit engineers from doing so because it's not intentional, the type has to be correct. if engineers would like to pass `any`, they don't need `MarkReadonly` type. Do you have any real-world examples where you would need to pass `any` as a second parameter?

som-sm:
@Beraliv Sorry, I'm not sure I fully understand.

But are you suggesting that we should prevent users from passing `any` as the second argument, something like this:

```ts
import { IsAny } from "ts-essentials";

export type MarkRequired<
  Type,
  Keys extends IsAny<NoInfer<Keys>> extends true ? never : keyof Type,
> = Type extends Type ? Type & Required<Omit<Type, Exclude<keyof Type, Keys>>> : never;

type T = MarkRequired<{ a: 1; b: 2 }, any>;
//                                    ~~~
//                                    Type 'any' does not satisfy the constraint 'never'
```

Playground link: https://tsplay.dev/WyZJ2w

If so, I'm fine with that. But, I believe if `any` is allowed, then it should behave correctly.

Beraliv:
Hey @som-sm! Thank you for your patience! Kindly ask to rebase your PR against the main branch. I'll return to you later this week

som-sm:
@Beraliv I've resolved the conflicts!

<br>

Also, gave this a thought once again, and I think we should handle the `any` case (_instead of preventing the `any` case), because:
- There’s no built-in utility to make all keys writable, so allowing `any` as the second argument lets `Mark-Writable` serve this purpose.
- Additionally, both Mark-Optional and Mark-Readonly already treat any as an instruction to mark all keys as optional or readonly, respectively.

Beraliv (approved):
@som-sm thank you for your patience! I just re-reviewed your PR and I agree with your approach: let's make all utilities consistent. Unit tests look great! Approving and merging your changes!

PR #418, opened 2024-11-03
Fix: Make `MarkOptional<Type, Keys extends keyof Type>` assignable to `Partial<Type>`

som-sm:
## PR Checklist

- [ ] Addresses an existing open issue: related to #000
- [X] Steps in [Contributing](https://github.com/ts-essentials/ts-essentials/blob/master/CONTRIBUTING.md) were taken

## Overview

Currently, generic instantiations of `MarkOptional` are not assignable to `Partial`. But I don't see a reason why it shouldn't.

image

<br />
<br />

This PR:
1. Refactors the implementation of `MarkOptional` to make it assignable to `Partial`.
2. Adds some more test cases for `MarkReadonly`.

PR #422, opened 2024-11-20
Feat: Prettify output of Mark-* & Merge types

som-sm:
## PR Checklist

## Overview

The output of utility types like `Mark-Optional` is currently not prettified, making it difficult to comprehend the resulting types.

image<br />

**This PR:**
- Prettifies output of `Mark-*` and `Merge` types.
   
  image
- Tests the output of types prettified types using the [`eslint-plugin-expect-type`](https://www.npmjs.com/package/eslint-plugin-expect-type) ESLint plugin.
- Adds `eslint` to `test` and `test:fix` scripts.

<br />

### Why wrap `Prettify` with `Extract` 🤔?

When we wrap `MarkOptional` with `Prettify`, TS is NOT convinced that `Prettify<MarkOptional<Type, Keys>` is just the same as `MarkOptional<Type, Keys>` because of which the [newly added assignability test for `MarkOptional`](https://github.com/ts-essentials/ts-essentials/pull/418/files#diff-0f36ef357fa0f9ba966a3b9d35853efbc6446a0a9c6dae039eaea4777e7bda54) fails.

By wrapping `Prettify` with `Extract<..., Type>`, we explicitly tell TS that the output of `Prettify` is always assignable to it's input. Here's [a playground](https://tsplay.dev/wXxJOw) with detailed explanation.

<br />

I couldn't find any known hacks for this `Prettify` issue, so this is something I've come up on my own, hence there might be room for improvement. LMK your thoughts on this.


lib/prettify/index.ts
```diff
@@ -1,5 +1,8 @@
 export type Prettify<Type> = Type extends Function
   ? Type
-  : {
-      [Key in keyof Type]: Type[Key];
-    } & {};
```
som-sm, line 5: Removed the additional intersection with `{}`, don't think it's necessary.

Beraliv:
Hey @som-sm! Thank you for your PR once again, I'll have a look at it this week!

lib/prettify/index.ts
Beraliv: `& {}` is required to avoid seeing `Prettify<...>` in a more nested examples. See more in https://stackoverflow.com/a/78465372. So I'll suggest reverting this change

package.json
```diff
@@ -12,8 +12,8 @@
     "build": "rimraf ./dist && tsc -p tsconfig.prod.json --outDir ./dist",
     "formatDeclarations": "prettier --ignore-path *.js --write dist/*.d.ts",
     "prepublishOnly": "yarn test && yarn build && yarn formatDeclarations",
-    "test": "node scripts/update-test-tsconfig.js && prettier -c {bench,lib,test}/**/*.ts && tsc --project tsconfig.test.json --outDir ./testDist && rimraf ./testDist",
-    "test:fix": "node scripts/update-test-tsconfig.js && prettier --write {bench,lib,test}/**/*.ts",
+    "test": "node scripts/update-test-tsconfig.js && prettier -c {bench,lib,test}/**/*.ts && eslint && tsc --project tsconfig.test.json --outDir ./testDist && rimraf ./testDist",
```
Beraliv, line 15: Feedback: `eslint` requires extra 15s (30s before & 45s now) for running test on each TypeScript version which isn't ideal

lib/mark-optional/index.ts
```diff
@@ -1,5 +1,6 @@
 import { OptionalKeys } from "../optional-keys";
+import { Prettify } from "../prettify";
 
 export type MarkOptional<Type, Keys extends keyof Type> = Type extends Type
-  ? Partial<Type> & Required<Omit<Type, Keys | OptionalKeys<Type>>>
+  ? Prettify<Partial<Type> & Required<Omit<Type, Keys | OptionalKeys<Type>>>>
```
Beraliv, line 5: I'd like to challenge your decision and ask what's the point of doing `Prettify` when your type is very complicated? On really big projects I don't think it makes much sense. It will be unreadable and cut in half. This type is available so user can manually apply it if it's useful, but I don't think we need to make this decision here and in other utility functions

lib/prettify/index.ts
```diff
@@ -1,5 +1,8 @@
...
+  : Extract<
```
Beraliv, line 3: Great optimisation on the assignability, thank you for giving a great example in Playground

Beraliv:
@som-sm a kind reminder to add changelog before ticking the PR checklist item "Steps in [Contributing](https://github.com/ts-essentials/ts-essentials/blob/master/CONTRIBUTING.md) were taken"

lib/prettify/index.ts
som-sm: The problem mentioned in https://stackoverflow.com/a/78465372 does happen when `Prettify` is _just_ a mapped type.

And it doesn't have to be nested; the problem shows up even with unions:

image

<br><br>

However, our implementation of `Prettify` is _not_ merely a mapped type, there's a conditional that wraps the mapped type, so it doesn't suffer from the same problem. And the display tests ensure that this behaviour remains intact and doesn't break.

image

image

<br><br>

Playground link: https://tsplay.dev/WYxgQm

package.json
som-sm: Umm...I tried a couple more times but it was never that slow, was mostly around 35-36s.

Anyways, I've moved the display tests to separate files, reducing the no. of files we've to lint. Not sure, if it actually does improve anything, but the execution time seems to be consistently around the 35s mark. Can you try running the job a couple of times to see how it does now?

image

<br><br>

> Are you aware of any other alternatives that have better execution time? 

No, not really!

som-sm: Umm...I tried a couple more times but it was never that slow, was mostly around 37-38s.

Anyways, I've moved the display tests to separate files, reducing the no. of files we've to lint. This should slightly improve the execution time, it now seems to be consistently around the 35s mark. Can you try running the job a couple of times to see how it does now?

image

<br><br>

> Are you aware of any other alternatives that have better execution time? 

No, not really!

lib/mark-optional/index.ts
som-sm: I still believe there’s value in using `Prettify` with types like `Mark*`, simply because it doesn’t degrade the experience—at worst, it’s neutral (e.g., when the display gets trimmed), and at best, it improves readability.

Btw, `type-fest` also prettifies [their version of `Mark*` types](https://github.com/sindresorhus/type-fest/blob/main/source/set-optional.d.ts#L30).

@Beraliv Do you see any potential downsides of using `Prettify` here?

som-sm:
@Beraliv Apologies, I missed the changeset 🙏🏻, I've added it now.

I’ve also responded back to all your questions—let me know your thoughts.

lib/prettify/index.ts
Beraliv: Okay, I extended your checks - https://tsplay.dev/WKQoyN. It consists of (1) conditional type + mapped type, (2) mapped types with NO intersection and (3) mapped types WITH intersection. Only second solution resolves `Array<SimplePrettify<A>>` as `SimplePrettify<A>[]`

Beraliv: I don't see any other reasons to keep `{}` so I'm happy to remove it for now

lib/mark-optional/index.ts
Beraliv: @som-sm agree to disagree, but let's find out about it. Worst case scenario - it won't be useful, it will break something serious so we will roll it back, document that X didn't work out because of (1), (2), etc. So I'm happy for you to keep `Prettify`

package.json
Beraliv: I'm comparing yarn test run across all TypeScript versions in master branch (7s worst) vs this branch (13s worst), so the time difference is 6s by just running `yarn test` but I assume you have to download all dependencies that may take extra 5-7s

Beraliv: Comparing usage: [your previous branch](https://github.com/ts-essentials/ts-essentials/actions/runs/12182991270/usage) 6m 9s vs 7m 12s [this branch](https://github.com/ts-essentials/ts-essentials/actions/runs/12194704372/usage), so it's over by a minute

I would say that the benefit is still worth it, so let's keep it

Beraliv (approved):
LGTM! Great work!

PR #424, opened 2024-12-02
Fix: `Tail` with optional, readonly and non-tuple arrays

som-sm:
## PR Checklist

## Overview

I came across [this issue](https://github.com/sindresorhus/type-fest/issues/978) in `type-fest` regarding their implementation of `Tail`, which doesn't work as expected when instantiated with an array of all optional elements. Our implementation of `Tail` also fails in this case; when instantiated with `[1?, 2?, 3?]`, it returns `never` instead of the expected `[2?, 3?]`.
<br>
Additionally, there are two other issues with the current implementation:
1. **Readonly Arrays**: `Tail` incorrectly returns `never` when instantiated with `readonly` arrays.
2. **Non-Tuple Arrays**: When applied to non-tuple arrays (e.g., `string[]`), `Tail` currently returns `never`. I believe it should instead return back the same type, e.g., `Tail<string[]>` should return `string[]`. However, in such scenarios, `type-fest` also returns an empty array, which is kinda equivalent to our version of `never`, though I'm unsure why.

@Beraliv If you think we should continue to return `never` for non-tuple arrays, then the [solution I proposed here](https://github.com/sindresorhus/type-fest/pull/977#issuecomment-2509125871) should work.

### DIFF
--- a/lib/tail/index.ts
+++ b/lib/tail/index.ts
@@ -1,3 +1,7 @@
 import { AnyArray } from "../any-array";
 
-export type Tail<Type extends AnyArray> = Type extends [any, ...infer Rest] ? Rest : never;
+export type Tail<Type extends AnyArray> = Type extends readonly []
+  ? never
+  : Type extends readonly [any?, ...infer Rest]
+  ? Rest
+  : never;
--- a/test/index.ts
+++ b/test/index.ts
@@ -625,15 +625,29 @@ function testNoop() {
   const callback: (_: number) => void = noop;
 }
 
-function testHeadTail() {
+function testHead() {
   type List1 = [number, string, boolean, "a" | "b"];
 
   type TestHead = Assert<IsExact<Head<List1>, number>>;
   type TestHeadOnEmpty = Assert<IsExact<Head<[]>, never>>;
+}
 
-  type TestTail = Assert<IsExact<Tail<List1>, [string, boolean, "a" | "b"]>>;
-  type TestTailOnEmpty = Assert<IsExact<Tail<[]>, never>>;
-  type TestTailOn1Elem = Assert<IsExact<Tail<[number]>, []>>;
+function testTail() {
+  type cases = [
+    Assert<IsExact<Tail<[]>, never>>,
+    Assert<IsExact<Tail<readonly []>, never>>,
+    Assert<IsExact<Tail<[1, 2, 3 | 4]>, [2, 3 | 4]>>,
+    Assert<IsExact<Tail<readonly [1, 2, 3 | 4]>, [2, 3 | 4]>>,
+    Assert<IsExact<Tail<[1]>, []>>,
+    Assert<IsExact<Tail<[1?]>, []>>,
+    Assert<IsExact<Tail<[1, 2] | [3, 4]>, [2] | [4]>>,
+    Assert<IsExact<Tail<[1?, 2?, 3?]>, [2?, 3?]>>,
+    Assert<IsExact<Tail<[1, 2, ...number[]]>, [2, ...number[]]>>,
+    Assert<IsExact<Tail<number[]>, number[]>>,
+    Assert<IsExact<Tail<readonly number[]>, number[]>>,
+    Assert<IsExact<Tail<any>, unknown[]>>,
+    Assert<IsExact<Tail<never>, never>>,
+  ];
 }
 
 function testExact() {


test/index.ts
```diff
@@ -625,15 +625,29 @@ function testNoop() {
...
+    Assert<IsExact<Tail<[1, 2, 3 | 4]>, [2, 3 | 4]>>,
+    Assert<IsExact<Tail<readonly [1, 2, 3 | 4]>, [2, 3 | 4]>>,
+    Assert<IsExact<Tail<[1]>, []>>,
+    Assert<IsExact<Tail<[1?]>, []>>,
+    Assert<IsExact<Tail<[1, 2] | [3, 4]>, [2] | [4]>>,
+    Assert<IsExact<Tail<[1?, 2?, 3?]>, [2?, 3?]>>,
+    Assert<IsExact<Tail<[1, 2, ...number[]]>, [2, ...number[]]>>,
+    Assert<IsExact<Tail<number[]>, number[]>>,
+    Assert<IsExact<Tail<readonly number[]>, number[]>>,
+    Assert<IsExact<Tail<any>, unknown[]>>,
```
Beraliv, line 648: (optional) this one is interesting, do you know why it equals to `unknown[]`? If you can document it above in the comment, it would be great

lib/tail/index.ts
```diff
@@ -1,3 +1,7 @@
 import { AnyArray } from "../any-array";
```
Beraliv, line 1: please add the changelog with the clear description

test/index.ts
som-sm: @Beraliv If you think this makes sense, then I'll add a link to this next to the test case.

lib/tail/index.ts
som-sm: Added the changeset in [this commit](https://github.com/ts-essentials/ts-essentials/pull/424/commits/b1213c5e6788571b98f3783d1d2390aca437a805), please verify once.

Beraliv: Great work, LGTM!

test/index.ts
Beraliv: Right, thank you for clarifying it for me!

> Now, Rest happens to be unknown[] probably because in the conditional Type extends readonly [any?, ...infer Rest], when Type is any, which is not an array, Rest falls back to its default constraint of unknown[]. A similar thing happens if Type is [].

This makes total sense, so it's a conditional type with inference that makes it `unknown[]`

> If we add an explicit constraint of any[] to Rest, then we'd get back any[] for any.

Please keep it as it is (because conditional constraint are not supported in all TypeScript versions that we support)

Beraliv: > @Beraliv If you think this makes sense, then I'll add a link to this next to the test case.

Please do!

Beraliv:
@som-sm please finalise your changes and I'll be happy to get it merged

lib/tail/index.ts
```diff
@@ -1,3 +1,7 @@
...
 
-export type Tail<Type extends AnyArray> = Type extends [any, ...infer Rest] ? Rest : never;
+export type Tail<Type extends AnyArray> = Type extends readonly []
+  ? never
+  : Type extends readonly [any?, ...infer Rest]
```
Beraliv, line 5: QQ: is it important that we use `any?` and not `any`?

I wouldn't see the difference if you asked me, because `any` is a top type, so from my understanding, leaving `[any, ...]` should work fine as well

If it doesn't, my bad, happy to accept it as is

Beraliv (approved):

lib/tail/index.ts
som-sm: > QQ: is it important that we use any? and not simply any?

Yes, the optional modifier is actually necessary, it doesn't work without that.

<br>

Consider the example of `[string?]` and `[any]`. The tuple `[string?]` doesn't extend `[any]` because `[any]` defines a tuple with exactly one element, however, `[string?]` allows for the possibility of an empty array.

```ts
type Test1 = [string?] extends [any] ? true : false; // Without optional
//   ^? type Test1 = false

type Test2 = [string?] extends [any?] ? true : false; // With optional
//   ^? type Test2 = true
```

This is similar to how an optional property wouldn't extend a non-optional `any` property:
```ts
type Test1 = { a?: string } extends { a: any } ? true : false; // Without optional
//   ^? type Test1 = false

type Test2 = { a?: string } extends { a?: any } ? true : false; // With optional
//   ^? type Test2 = true
```

Playground: https://tsplay.dev/WPQrYw

test/index.ts
som-sm: Added it in [this commit](https://github.com/ts-essentials/ts-essentials/pull/424/commits/3ae6fdf5eaedaa32584e3f56446a1c5886791fb4). Also, added a test case for `any[]`.

som-sm: Also, included test cases for union of tuples with different lengths ([#e417a8c](https://github.com/ts-essentials/ts-essentials/pull/424/commits/e417a8c93ae2b675d5283b8e8b6df0ca98b92c9d)).

Because, this is the case (`[1, 2] | []`) where `Head` type fails, so it's worth having it for `Tail` as well.

lib/tail/index.ts
Beraliv: Great, it's clear now! Thanks

PR #436, opened 2025-06-12
Fix: `IsTuple<Type>` when `Type` is an array containing rest element

som-sm:
## PR Checklist

## Overview

Currently, `IsTuple` doesn’t recognise arrays with a rest element and optional elements as tuples, such as `[number?, …string[]]`.
```ts
import { IsTuple } from 'ts-essentials';

type T = IsTuple<[number?, ...string[]]>;
//   ^? type T = never
```

This PR fixes that by updating the implementation of `IsTuple`.

<hr>

However, the above fix causes the following test to fail in `DeepNullable` (similar with `DeepUndefinable`):
```ts
Assert<IsExact<DeepNullable<never[]>, null[]>>
```

There are two problems here:
1. **Incorrect behavior in `IsTuple` for `never[]`:**
   Previously, `IsTuple<never[]>` returned `never[]`, which is incorrect—it should return `never`. The updated implementation corrects this.
   
2. **Incorrect handling of `never` in `DeepNullable`:**
   Once `IsTuple<never[]>` correctly returns `never`, `DeepNullable<never[]>` starts returning `never[]` instead of `null[]`. This happens because `DeepNullable<never>` currently returns `never`.
   
   The issue becomes more apparent when testing with a case that doesn't involve `IsTuple`:
   ```ts
   type T = DeepNullable<{ foo: never }>;
   //    ^? { foo: never } (expected: { foo: null })
   ```
   
   To fix this, this PR makes `DeepNullable<never>` return `null`.


test/deep-readonly.ts
```diff
@@ -63,7 +65,6 @@ function testDeepReadonly() {
       >
     >,
     Assert<IsExact<DeepReadonly<ComplexNestedRequired>, ComplexNestedReadonly>>,
-    Assert<IsExact<DeepReadonly<{ foo: [string, ...number[]] }>, { readonly foo: readonly [string, ...number[]] }>>,
```
som-sm, line 66: Just moved this case above along with other array cases.

Beraliv (approved):
Great work @som-sm! Thank you for raising this PR

Issue #468, opened 2026-04-09
Bug: Incorrect DeepPartial mapping for variadic tuples results in ...undefined[] rest elements


nathanbowang:
### Versions

| package                                | version |
| -------------------------------------- | ------- |
| `typescript`                           | `5.9.3` |
| `ts-essentials`                        | `10.0.3` |

### Repro Code

```typescript
import {DeepPartial} from 'ts-essentials'

type A = [string, ...string[]]
type B = DeepPartial<A>

// Type '[string, string]' is not assignable to type '[(string | undefined)?, ...undefined[]]'.
//  Type at position 1 in source is not compatible with type at position 1 in target.
//    Type 'string' is not assignable to type 'undefined'.ts(2322)
const b: B = ['a', 'b']
```

### TypeScript Playground

https://www.typescriptlang.org/play/?#code/JYWwDg9gTgLgBAbwCIFMVgAoENbCwGwF84AzKCEOAchgGcBaFW2lAOxj31qoCgeYAnmBRwAgnAC8cANq0YUYKwDmAGjgA6TXIXLpAXT38hIgEKS4qdNlwEAPKIB8fAMYRWcuACMAXHDNTpKiwqNSpPKkMgA

### Expected Result

B is [(string | undefined)?, ...(string | undefined)[]]

### Actual Result

When applying DeepPartial to a variadic tuple like [string, ...string[]], the rest elements are incorrectly mapped to undefined[] instead of (string | undefined)[]. This causes type errors when assigning standard arrays (e.g., ['a', 'b']) because subsequent elements are expected to be undefined.

Beraliv:
This will be released as part of https://github.com/ts-essentials/ts-essentials/pull/438 (v10.2.0)

Beraliv:
@nathanbowang thank you for raising the issue, @som-sm thanks for the quick fix!

Beraliv:
Released today under v10.2.0

Issue #434, opened 2025-06-11
Bug: [DeepReadonly] converts rest element type in tuple to never


radoslavkarlik:
### Versions

| package                                | version |
| -------------------------------------- | ------- |
| `typescript`                           | `5.8.3` |
| `ts-essentials`                        | `10.1.0` |

### Repro Code

```typescript
type Result = DeepReadonly<[number, ...number[]]>;
```

### TypeScript Playground

https://www.typescriptlang.org/play/?#code/JYWwDg9gTgLgBAbzgEQKarAJVQQwCYQB2ANgJ5wC+cAZlBCHAOQwDOAtKiy6oTMDsRaMAsACgxqAB6RYcGKTCo42FgFdi8ALwp0WXARKkAPAG1CqkACNUUADRwAdE-NWbJgLruAfAG4gA

### Expected Result

readonly [number, ...number[]]

### Actual Result

readonly [number, ...never[]]

Beraliv:
From what I see, TypeScript >=5.4 versions have been impacted

* 5.4.5 - https://tsplay.dev/mbbq9m (returns `readonly [number, ...never[]]`)
* 5.3.3 - https://tsplay.dev/mpYb7W (returns `readonly [number, ...number[]]`)

Beraliv:
@som-sm amazing, thank you for confirming my concerns, you made great investigation!

So we either remove IsTuple completely (I need to double check if this is acceptable for other Deep utility types). If we still need to keep it, let's move it to the left side to avoid the bug.

Happy either way!

If you have capacity, you can raise the PR, otherwise I'll land it in the evening, when I'm back from work.

Beraliv:
@som-sm thank you again for raising the PR so quickly! The change is perfect, I've merged it. It was released as part of https://github.com/ts-essentials/ts-essentials/pull/437, [v10.1.1](https://github.com/ts-essentials/ts-essentials/releases/tag/v10.1.1)

PR #469, opened 2026-04-10
Fix: `DeepPartial<Type>` when `Type` is an array containing rest element

som-sm:
## PR Checklist

- [X] Addresses an existing open issue: related to #468

## Overview

This is the same issue as mentioned in [this comment](https://github.com/ts-essentials/ts-essentials/issues/434#issuecomment-2966878618):
> The issue is indeed with the `Type extends IsTuple<Type>` conditional, here’s a minimal reproducible example for the same:
> 
> import { IsTuple } from 'ts-essentials';
> 
> // no-op mapped type with `IsTuple` conditional
> type Test<T> = T extends IsTuple<T> ? { [P in keyof T]: T[P] } : never;
> 
> type T = Test<[bigint, ...string[]]>; // `...string[]` changes to `...never[]`
> //   ^? type T = [bigint, ...never[]]
> [Playground](https://www.typescriptlang.org/play/?#code/JYWwDg9gTgLgBAbzgSQM4BUCuYA2BTOAXzgDMoIQ4ByGVAWj1VTwDsZgBDHVKgbgFgAUEID0IuCwh0IYOCA5gweACZwYATyVwA7sBgALOAAM0WXHiNwAxhBbK9wW1yEat6RjAA86AHxwAvHDocHgAHjCsyqgoGNj43n4A-IhwANoACnDALHAA1njqECRBALoAXEEZJURwFSx4AG54UALCgq4EwYHuqF6pAEbAAObZMAA0cAB0071Q2UOpJSU+vHBixtOTs-OLllb6HCxDjGoQG9P1TVC7ouJwcAB6yR1BAWmDI2wTm5fNiyVAA)
> 
> So, `T` ends up being `[bigint, ...never[]]`, which is incorrect—it should’ve preserved the `...string[]`.

<br>

This PR _simply_ removes the explicit handling for arrays, as they can be handled the same way as objects without any issues. So, the following _homomorphic_ mapped type should be able to handle both arrays and objects:
https://github.com/ts-essentials/ts-essentials/blob/bb7d19fe4754a6809267c2ba355500f0ea386573/lib/deep-partial/index.ts#L28

### DIFF
--- a/.changeset/fuzzy-nails-sip.md
+++ b/.changeset/fuzzy-nails-sip.md
@@ -1,0 +1,5 @@
+---
+"ts-essentials": patch
+---
+
+Fix `DeepPartial<Type>` when `Type` is an array containing rest element, like `[string, ...number[]]`
--- a/lib/deep-partial/index.test.ts
+++ b/lib/deep-partial/index.test.ts
@@ -53,6 +53,10 @@ function testDeepPartial() {
     Assert<IsExact<DeepPartial<readonly never[]>, readonly undefined[]>>,
     Assert<IsExact<DeepPartial<[1, 2, 3]>, [(1 | undefined)?, (2 | undefined)?, (3 | undefined)?]>>,
     Assert<IsExact<DeepPartial<readonly [1, 2, 3]>, readonly [(1 | undefined)?, (2 | undefined)?, (3 | undefined)?]>>,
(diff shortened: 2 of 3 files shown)


nathanbowang (approved):
LGTM, Thanks

Beraliv:
Thanks @som-sm! Acknowledged, will review next week

Beraliv (approved):
thanks! the changes make sense, approving and merging!

